PR #28, opened 2018-10-05
Move tesp metadata code & tests here

jeremyh:
The galpgs repository now needs to use tesp's metadata generation code. But tesp has a lot of difficult dependencies to install (all of wagl: which has Fortran compilation, stricter rasterio and dependency requirements etc). The vast majority of tesp's code is irrelevant to galpgs. 

A better option is to move tesp's metadata gen code here. Both galpgs and tesp already depend on eodatasets, and eodatasets' role already involves metadata generation.

This also replaces pylint (which eodatasets was using) with flake8 (which tesp is using). The eodatasets build is currently failing due to newer pylint releases, and it seems easier to switch to the more pragmatic flake8.


jeremyh:
Reopening on local branch to avoid aws-access-key issue with Travis.

PR #30, opened 2018-10-09
Added deterministic uuid for ls and scihub prepare scripts; created an entrypoint

ASVincent:
* Fix a deterministic UUID for the level 1 datasets (based on a fixed namespace and package name).
* Create an entrypoint for the prepare scripts in eod
* Enhance docstrings to improve click cli descriptions
* Added tests_require to simplify running tests in constrained environments (NCI)

### DIFF
--- a/eodatasets/prepare/ls_usgs_l1_prepare.py
+++ b/eodatasets/prepare/ls_usgs_l1_prepare.py
@@ -216,9 +216,11 @@ def prepare_dataset(path):
         mtl_filename
     )
 
+    product_id = mtl_doc['metadata_file_info']['landsat_product_id']
+
     # Generate a deterministic UUID for the level 1 dataset
     return {
-        'id': str(uuid.uuid5(USGS_UUID_NAMESPACE, path.name)),
+        'id': str(uuid.uuid5(USGS_UUID_NAMESPACE, product_id)),
         'product_type': 'level1',
         'extent': {
             'center_dt': '{} {}'.format(
--- a/eodatasets/prepare/ls_usgs_l1_prepare.py
+++ b/eodatasets/prepare/ls_usgs_l1_prepare.py
@@ -24,6 +24,9 @@ try:
 except ImportError:
     pass
 
+# Static namespace to generate uuids for datacube indexing
+USGS_UUID_NAMESPACE = uuid.UUID('276af61d-99f8-4aa3-b2fb-d7df68c5e28f')
+
 LANDSAT_8_BANDS = [
     ('1', 'coastal_aerosol'),
     ('2', 'blue'),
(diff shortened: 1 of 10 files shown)


eodatasets/prepare/ls_usgs_l1_prepare.py
```diff
@@ -213,8 +216,9 @@ def prepare_dataset(path):
         mtl_filename
     )
 
+    # Generate a deterministic UUID for the level 1 dataset
     return {
-        'id': str(uuid.uuid5(uuid.NAMESPACE_URL, path.as_posix())),
+        'id': str(uuid.uuid5(USGS_UUID_NAMESPACE, path.name)),
```
jeremyh: I'd prefer to make it explicit that we're using the landsat product id. Perhaps use the one from the mtl? `mtl→metadata_file_info→landsat_product_id` 

There's also an (admittedly tiny) chance that they've renamed the directory, or are processing it themselves.

eodatasets/prepare/ls_usgs_l1_prepare.py
```diff
@@ -284,7 +288,8 @@ def find_gz_mtl(ds_path, output_folder):
         members = [m for m in tar_gz.getmembers() if mtl_pattern.search(m.name)]
         if not members:
             return None
-        tar_gz.extractall(output_folder, members)
+        # extractall requires the output location as a string
```
jeremyh, line 293: I'm surprised the tests didn't fail -- they extract a tar. Different Python version?

The comment probably isn't necessary for future readers: we do the same str/path conversion in lots of places (it could be relevant in the commit message though, if this fix was alone).

eodatasets/scripts/genprepare.py
```diff
@@ -0,0 +1,16 @@
...
+
+@click.group()
+@click.version_option()
+def run():
+    pass
+
+
+run.add_command(s2_scihub, name='s2-cophub')
+run.add_command(ls_usgs, name='ls-usgs')
+run.add_command(s2_awspds, name='s2-awspds')
```
jeremyh, line 16: :+1: Much nicer than running the py files. 

We can probably later expand the integration tests to run through this layer too (but it doesn't have to be done now).

jeremyh (commented):
Nice work

eodatasets/prepare/ls_usgs_l1_prepare.py
jeremyh: The integration tests [currently ignore the uuid](https://github.com/GeoscienceAustralia/eo-datasets/blob/b44b7102cdf1a805db2d74d94a9c3a2f9913cd0b/tests/integration/prepare/test_prepare_ls5_l1.py#L18), but we could assert a specific one now the generation is stable.

eodatasets/prepare/ls_usgs_l1_prepare.py
ASVincent: I was testing it in the NCI environment; just using setup.py to install.

jeremyh (approved):

PR #32, opened 2018-10-15
Include size_bytes field in usgs l1 datasets

jeremyh:
- All of three of the tests were testing a tar file input, so I'm switching one to take an untarred directory (matching galpgs's usage).
- Includes a fix for the current Travis failure (AWS key being unset)


uchchwhash (approved):
More to process than I can handle at the moment. But looks good.

PR #35, opened 2019-01-14
Bugfix/processing time s2 pds

ASVincent:
* Fixes a mistake in the aws pds prepare script
* adds tests for s2a (safe archive), s2b (aws pds)
* adds normalise_nci_symlinks to landsat tests (to allow them to run on NCI)
* minor modification of the tests/integration/prepare/common file to place invoked script behind a variable.
* enable specifying independent datastrip metadata location


eodatasets/prepare/s2_l1c_aws_pds_prepare.py
```diff
@@ -96,6 +96,7 @@ def valid_region(images, mask_value=None):
 
 
 def _to_lists(x):
+    # type (Tuple) -> List
```
jeremyh: It needs a colon after "type" to be parsed as a type hint:
```suggestion
    # type: (Tuple) -> List
```

eodatasets/prepare/s2_l1c_aws_pds_prepare.py
```diff
@@ -137,6 +138,7 @@ def transform(p):
 
 
 def get_datastrip_info(path):
+    # type (Path) -> (Dict, str)
```
jeremyh:
```suggestion
    # type: (Path) -> (Dict, str)
```

eodatasets/prepare/s2_l1c_aws_pds_prepare.py
```diff
@@ -154,27 +156,35 @@ def get_datastrip_info(path):
             uuid.NAMESPACE_URL,
             datastrip_metadata + "#" + product_info['id']
         )
-    return datastrip_metadata, persisted_uuid
+    return (datastrip_metadata, persisted_uuid)
 
 
 def get_tile_info(path):
+    # type Path -> str
```
jeremyh:
```suggestion
    # type: (Path) -> str
```

eodatasets/prepare/s2_l1c_aws_pds_prepare.py
```diff
@@ -154,27 +156,35 @@ def get_datastrip_info(path):
...
     with (path / 'tileInfo.json').open() as fd:
         tile_info = json.load(fd)
-        tile_path = tile_info.get('path')
-    return tile_path
+        return tile_info.get('path')
 
 
-def prepare_dataset(path):
+def prepare_dataset(path, datastrip_path=None):
+    # type (Path, Optional(Path)) -> Dict
```
jeremyh: Returns a list of documents (PyCharm shows a warning about this)

```suggestion
    # type: (Path, Optional[Path]) -> List[Dict]
```


(And ideally ```from typing import Dict, Tuple, List, Optional``` at the top of the file for all of these docs)

tests/integration/prepare/test_prepare_s2a_l1c_safe.py
```diff
@@ -0,0 +1,344 @@
...
+                'PVI': {
+                    'layer': 1,
+                    'path': path_offset('QI_DATA/T56JMM_20180629T000241_PVI.jp2')
+                },
+                'TCI': {
+                    'layer': 1,
+                    'path': path_offset('IMG_DATA/T56JMM_20180629T000241_TCI.jp2')
+                }
+            },
+            'cloud_cover_percentage': '0.0',
```
jeremyh: It wasn't obvious until you added a test for it, but these properties were numbers in the previous collection rather than strings.

From the old tests: https://github.com/GeoscienceAustralia/eo-datasets/blob/1062c02160380b7ebbf2076b2c03ddb4f1bb32bf/tests/integration/test_ls8_nbar.py#L540-L543

I'm guessing this affects all of these newly-added prepare scripts. We can address this in a separate issue if you want. Many of the newly added fields in this section probably need the same change.

eodatasets/prepare/s2_l1c_aws_pds_prepare.py
```diff
@@ -356,6 +366,7 @@ def prepare_dataset(path):
 
 
 def absolutify_paths(doc, path):
+    # type (Dict, str) -> Dict
```
jeremyh:
```suggestion
    # type: (Dict, str) -> Dict
```

jeremyh (changes requested):
Great to see tests for the new stuff! Thanks Alex.

A few minor type hints to adjust. We don't currently have a linter set up to complain about these, so they're easy to miss.

ASVincent:
Pushed more edits converting text floats to floats, I need to run this on a sample at NCI before merging.

ASVincent:
I've applied changes (as requested), discussed changes with yourself and Josh (impacts) and run it against a sample of our collection ~ (200 packages).
Are you able to complete the review?
-- Cheers

jeremyh (approved):

PR #36, opened 2019-01-22
Add tar recompression tool

jeremyh:
For repacking USGS L1 files, as proposed and prototyped by @sixy6e 

- Includes a few commits that clean up python warnings (such as py3 deprecations).
- Includes tests that perform a repackage on both a`.tar.gz` and an unpackaged folder.

```
 $ eod-recompress-tar --help
Usage: eod-recompress-tar [OPTIONS] [PATHS]...

  Repackage a USGS Collection-1 tar for faster read access.

  They arrive as a *.tar.gz with inner uncompressed tiffs, which Josh's
  tests have found to be too slow to read.

  We compress the inner tiffs and store them in an uncompressed tar. This
  allows random reads within the files. We also append a checksum file at
  the end of the tar.

Options:
  --output-base DIRECTORY  The base output directory.  [required]
  --zlevel INTEGER RANGE   Deflate compression level.
  --block-size INTEGER     Compression block size (both x and y)
  --help                   Show this message and exit.

```


ASVincent (approved):
Looks great!

PR #37, opened 2019-03-12
Add cleanup-on-success option to repackager

jeremyh:
- Adds a `--clean-inputs` option which removes the original input dataset after successful recompress. (needed to avoid doubling the disk usage)
- Improves logging. Switched it to structlog.
- Add a `-f` option to read a list of inputs from a file (one per line)
- Various test improvements for edge cases

### DIFF
--- a/eodatasets/scripts/recompress.py
+++ b/eodatasets/scripts/recompress.py
@@ -369,9 +369,17 @@ def _recompress_image(
         output_dataset.update_tags(1, **input_image.tags(1))
 
 
+class PathPath(click.Path):
+    """A Click path argument that returns a pathlib Path, not a string"""
+    def convert(self, value, param, ctx):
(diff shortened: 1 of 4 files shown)


eodatasets/scripts/recompress.py
```diff
@@ -371,20 +404,43 @@ def main(paths: List[str], output_base: str, zlevel: int, block_size: int):
...
+        _LOG.info(
+            "node.finish",
+            host=socket.getfqdn(),
+            total_count=total,
+            failure_count=failures,
+        )
     sys.exit(failures)
 
 
+def please_remove(path: Path, excluding: Path):
```
sixy6e, line 428: so polite !!! :)

sixy6e (approved):
All looks good to me!

PR #40, opened 2019-04-09
Enabling old style of landsat ids

ASVincent:
* Enable fallback to landsat_scene_id for older datasets.

### DIFF
--- a/eodatasets/prepare/ls_usgs_l1_prepare.py
+++ b/eodatasets/prepare/ls_usgs_l1_prepare.py
@@ -274,7 +274,9 @@ def prepare_dataset_from_mtl(total_size: int,
         mtl_filename,
     )
 
-    product_id = mtl_doc['metadata_file_info']['landsat_product_id']
+    product_id = mtl_doc['metadata_file_info'].get(
+            'landsat_product_id',
+            mtl_doc['metadata_file_info']['landsat_scene_id'])
 
     additional = additional_props or {}
 


uchchwhash (dismissed):
:+1:

jeremyh:
I had thought scene ids were not unique enough: do they change when USGS reprocesses?

ASVincent:
@jeremyh you know more about this domain than I do, I saw in a previous rendition of a prepare script that it tried one and then the other, I assumed this was due to an evolving metadata specification (as such a justification had been provided elsewhere).
In your experience have you come across packages that are lacking the product_id field from the MTL file?

ASVincent:
Found a source describing the difference between Scene ID and product ID. 
Scene ID was an attribute of the pre-collection datasets, and based on the documentation is unique for a 
spatio-temporal observation and processing version.

It's not guaranteed to be safe (for deterministic UUID generation) but should be appropriate given that the plan is to move away from pre-collection data.
Reference File: https://landsat.usgs.gov/sites/default/files/documents/LSDS-833_V6.0.pdf

jeremyh:
Yeah, all the old datasets have it (but not for UUID generation), The existing/old eodatasets commands (`eod-generate-metadata`/`eod-package`) do our pre-collection packaging at the moment with some integration tests (but they're testing GA-processed examples, not USGS).

I'm worried the rest of this script may not work with pre-collection as they've changed multiple fields in the metadata, some of which could make it fall over if missing. If we want to support pre-collection MTLs in this script I'd prefer we added another integration test case for it.

PR #42, opened 2019-04-17
Adding prepare scripts for [MCD43A1,NCEP/NCAR reanalysis pr_wtr]

ASVincent:
Adds prepare scripts for the MCD43A1 datasets and NCEP/NCAR surface reanalysis 1 pr_wtr.eatm datasets.

TODO:
- ~Update prepare scripts with a description field -- possibly a DOI or link to CMI~
- [x] Confirm netCDF format for datacube
- ~confirm product_name, product_type fields with product owner~
- [x] Confirm product href for new metadata standard

Future Tasks
- provide downsampled MCD43A1 package for integration testing

Additional Files:
-  ncep resampler (to generate a reduced size (temporal) pr_wtr...nc file if it needs to be redone.


eodatasets/prepare/ncep_reanalysis_surface_pr_wtr.py
```diff
@@ -0,0 +1,166 @@
...
+def get_uuid(collection, idx):
+    """
+    Returns a deterministic uuid based on band index and gdal checksum
+    """
+    origin = collection.tags()['NC_GLOBAL#References']
+
+    return uuid.uuid5(
+        NOAA_WATER_VAPOUR_NS,
+        '{}?{}'.format(origin, urllib.parse.urlencode({
+            'checksum': collection.checksum(idx),
```
jeremyh: I think the gdal checksum is a 16-bit checksum intended for error-detection, right? How many of these datasets could we theoretically face?

(from dumping the example, I'm _guessing_ the other two fields are product-specific, not dataset-specific, so it's just the checksum that matters)

eodatasets/prepare/ncep_reanalysis_surface_pr_wtr.py
```diff
@@ -0,0 +1,166 @@
...
+            xkey: collection.bounds.left
+        },
+        'ur': {
+            ykey: collection.bounds.top,
+            xkey: collection.bounds.right
+        }
+    }
+
+
+def get_uuid(collection, idx):
```
jeremyh: These parameters were unclear to me without finding usages of the method. Maybe rename or add type decorators, or both?
```suggestion
def get_uuid(dataset: DatasetReader, band_index: int):
```

(I'm biased towards types because they make many tools more useful...)

eodatasets/prepare/ncep_reanalysis_surface_pr_wtr.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+
+    with rasterio.open(str(dataset), 'r') as collection:
+        collection_start_date = (
+            datetime.datetime.strptime(
+                collection.tags()['time#units'],
+                'hours since %Y-%m-%d %H:%M:%S.%f'
+            )
+        )
+        creation_dt = (
```
jeremyh: defined already a few lines above ([75](https://github.com/GeoscienceAustralia/eo-datasets/pull/42/files#diff-e4f346d610f818ad7bd62ee282ef5b18R75))

(pycharm is warning about this)

eodatasets/prepare/ncep_reanalysis_surface_pr_wtr.py
```diff
@@ -0,0 +1,166 @@
...
+                'center_dt': ds_dt.isoformat(),
+                'coord': get_coords(collection)
+            }
+            md['grid_spatial'] = {
+                'projection': {
+                    'geo_ref_points': get_coords(collection, xkey='x', ykey='y'),
+                    'spatial_reference': rasterio.crs.CRS(init='EPSG:4236').wkt
+                }
+            }
+            md['format'] = {'name': 'netCDF'}
```
jeremyh: Most of the dea products use capitalisation "NetCDF" from the looks of it: https://github.com/GeoscienceAustralia/digitalearthau/search?q=netcdf&unscoped_q=netcdf

(though humorously there's deviations in other repositories :smile: 
netcdf-variations)

eodatasets/prepare/ncep_reanalysis_surface_pr_wtr.py
```diff
@@ -0,0 +1,166 @@
...
+    return datasets
+
+
+def _process_datasets(output_dir, datasets):
+    """
+    Wrapper function for processing multiple datasets
+    """
+    for dataset in datasets:
+        docs = process_datasets(dataset)
+        outfile = output_dir / (dataset.stem + '-metadata.yaml')
```
jeremyh: We previously used `.ga-md.yaml` extension for sidecar metadata (there's a few codebases like sync which will check this: eg. https://github.com/GeoscienceAustralia/digitalearthau/blob/901a2988cadcdd91c4cbcc14105928b94dcfba09/digitalearthau/test_paths.py#L64

tests/integration/prepare/test_ncep_reanalysis_surface_pr_wtr.py
```diff
@@ -0,0 +1,663 @@
...
+                            "x": -1.25,
+                            "y": 91.25
+                        },
+                        "ur": {
+                            "x": 358.75,
+                            "y": 91.25
+                        }
+                    },
+                    "spatial_reference": (
+                        'GEOGCS["Hu Tzu Shan 1950",DATUM["Hu_Tzu_Shan_1950",SPHEROID'
```
jeremyh: I strongly prefer just the epsg "epsg:4236" rather than the full WKT, unless there's a need?

tests/integration/prepare/test_ncep_reanalysis_surface_pr_wtr.py
```diff
@@ -0,0 +1,663 @@
...
+        ncep_reanalysis_surface_pr_wtr.main,
+        '--output', str(output_path), str(NCEP_PR_WTR_FILE)
+    )
+
+    assert expected_metadata_path.exists()
+    docs = list(yaml.safe_load_all(expected_metadata_path.open()))
+
+    for idx in range(len(expected_doc)):
+        doc_diff = _diff(expected_doc[idx], docs[idx], exclude_paths="root['creation_dt']")
+        assert doc_diff == {}, pformat(doc_diff)
```
jeremyh: :+1: Nice tests.

(If it becomes unwieldy: just testing the first two docs and the total count would be fine with me → less code to keep up to date)

eodatasets/prepare/modis_usgs_mcd43a1.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    Generates a metadata document for each tile provided,
+    requires a path to the input tile (hdf) and the
+    corresponding xml document describing the dataset.
+    """
+    band_info = get_band_info(input_path)
+    xml_md = parse_xml(xml_file)
+
+    md = {}
+    md['id'] = uuid.uuid5(MCD43A1_NS, xml_md.get('granule_id', input_path.name))
```
jeremyh: I don't know anything about granule ids, but I presume they include processing time or proc version? (there's no example anywhere I can see in the diff or tests)

Fallbacks like `input_path.name` have been dangerous in the past (unexpected human usage of a script). Not sure about it.

tests/integration/prepare/test_ncep_reanalysis_surface_pr_wtr.py
```diff
@@ -0,0 +1,663 @@
...
+                    }
+                }
+            },
+            "product_name": "ncep_reanalysis_surface_pr_wtr",
+            "product_type": "auxiliary",
+            "sources": {}
+        },
+        {
+            "creation_dt": "2019-04-17T01:14:42.954377+00:00",
+            "description": "",
```
jeremyh: Exclude the description if blank.

jeremyh (commented):
Nice work Alex

eodatasets/prepare/ncep_reanalysis_surface_pr_wtr.py
ASVincent: origin is product specific, which would facilitate generating a different UUID if the source of the data changes, that could be done by changing the namespace. Not sure if your preference is to drop the parameter or not.
If I keep the origin (url) I should probably add the filename as I don't think I've encoded that in the UUID generation, or use the observation date.
The band_id * 6hrs provides the offset of the observation from the start of the year, I think I should use that with the checksum to distinguish between observations in the UUID generation.

tests/integration/prepare/test_ncep_reanalysis_surface_pr_wtr.py
ASVincent: description is to be filled in with a DOI or link to CMI.

eodatasets/prepare/modis_usgs_mcd43a1.py
ASVincent: I'll remove the fallback :)

tests/integration/prepare/test_ncep_reanalysis_surface_pr_wtr.py
ASVincent: Mostly I just wanted to get the tests over multiple days; not sure if this is an invalid desire.

eodatasets/prepare/ncep_reanalysis_surface_pr_wtr.py
ASVincent: Dataset count is roughly 1460 per annum, 60 year collection.

tests/integration/prepare/test_ncep_reanalysis_surface_pr_wtr.py
sixy6e: Here is a CMI link to the precipitable water content.

https://cmi.ga.gov.au/node/373

sixy6e (approved):
Great work!

PR #43, opened 2019-06-12
Fix spelling error in new prepare scripts; add them to cli

ASVincent:
* Adds new prepare scripts to eod-prepare
* Fixes a spelling error in new prepare scripts; used a common enum to provided consistent spelling


jeremyh (approved):
:+1:

Issue #48, opened 2019-07-22
Replace gdal-cli calls with pure-python/rasterio


jeremyh:
The cog-writing methods are descended from wagl's code, and use a mixture of both rasterio and the gdal shell commands (`gdal_translate` and `gdalwarp`).

This is not ideal.
- there's nothing guaranteeing the user has gdal on their path, as @omad has pointed out.
- it leads to "double" the memory usage, as the Python process grows its memory usage during rio operations, but this memory is not reused when calling the external commands. (this affected some pbs jobs)

jeremyh:
All are converted to rio now.

The thumbnail quality has gone up too, as the old code was doing (lossy) jpeg-conversion multiple times and had larger jpeg-artifacts.

PR #50, opened 2019-07-29
Remove the remaining gdal cli calls from the image writer

jeremyh:
It's purely rio-based now, and removes the dependency on scikit-image.

There's further room for improvement, such as avoiding keeping whole bands in memory, but that was already the case in the previous code.

### DIFF
--- a/eodatasets3/images.py
+++ b/eodatasets3/images.py
@@ -659,7 +659,7 @@ def _write_quicklook(
         *input_geobox.bounds,
     )
     reproj_grid = GridSpec(
-        (reprojected_width, reprojected_height), reprojected_transform, crs=out_crs
+        (reprojected_height, reprojected_width), reprojected_transform, crs=out_crs
     )
     # Calculate combined nodata mask
     nulls = numpy.zeros(input_geobox.shape, dtype="bool")
--- a/eodatasets3/images.py
+++ b/eodatasets3/images.py
@@ -568,52 +568,45 @@ class FileWrite:
         input_geobox: GridSpec,
         out: Path,
         out_scale=10,
+        # TODO: infer source range?
         src_range=(1, 3500),
         resampling=Resampling.bilinear,
     ):
         """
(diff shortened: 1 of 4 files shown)


jeremyh:
See #48

eodatasets3/images.py
```diff
@@ -559,157 +562,174 @@ def write_from_ndarray(
...
+                    num_threads=2,
+                )
+                del rescaled_data
+                ql_ds.write(reprojected_data, band_no)
+                del reprojected_data
+    del nulls
+    return reproj_grid
+
+
+def rescale_intensity(
```
sixy6e, line 714: Fine for the time being this is ok, but anything more advanced than a simple linear stretch, we can write it ourselves, or make use of an external library that provides it.

eodatasets3/images.py
```diff
@@ -559,157 +562,174 @@ def write_from_ndarray(
...
-                out_ds.write(scaled, band)
+    (and it saves us bringing in the entire dependency for one small method)
+    """
+    imin, imax = in_range
+    omin, omax = out_range
+
+    np.clip(image, imin, imax, out=image)
+    image -= imin
+    # We allocate a float array here, as with the original function.
+    # (I'm not sure how bad int truncation would affect us?)
```
sixy6e, line 731: I've tested out_ranges of [1, 255] and [0, 255] and they don't have any truncation side effects.  But my interests here are purely for display, as such the range [0, 255] or anything in-between is all I'm caring about.  Anything outside this range shouldn't have an issue either, as it would be following the same stretch logic.  Or we could include a set check of the range(0, 255) to ensure an input exists only within that range.

sixy6e (approved):
Looks good.  A few comments regarding the stretch, but largely depends on what we'll do in future anyway.

eodatasets3/images.py
jeremyh: Cool. We'll cross that bridge when it's needed.

PR #52, opened 2019-08-01
Add support for writing measurements from xarray.Datasets

omad:
Also fix a warning that pre-commit was throwing.

I don't have any automated tests yet. What's the state of testing with eodatasets at the moment?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
       hooks:
           - id: black
     - repo: https://github.com/adrienverge/yamllint.git
-      sha: v1.15.0
+      rev: v1.16.0
       hooks:
           - id: yamllint
     - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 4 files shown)


jeremyh:
Great, thanks!

The test situation is ok at the moment -- all the public methods have a bit of coverage. The [assemble tests](https://github.com/GeoscienceAustralia/eo-datasets/blob/eodatasets3/tests/integration/test_assemble.py) do some basics, but most of it is indirect through the (more thorough) [wagl](https://github.com/GeoscienceAustralia/eo-datasets/blob/eodatasets3/tests/integration/test_packagewagl.py) and [level1](https://github.com/GeoscienceAustralia/eo-datasets/blob/eodatasets3/tests/integration/prepare/test_prepare_landsat_l1.py) tests. There's a few standalone unit tests too.

One option may be to change [the minimal package test](https://github.com/GeoscienceAustralia/eo-datasets/blob/44bdaf44c2673ab7421bb029117341d9f7055979/tests/integration/test_assemble.py#L171) to write an xarray instead of a file. It's using a write call that's well covered already. Or we add a separate test.

omad:
I think the biggest annoyance with testing this is that it relies on *datacube-core* CRS and GeoBox functionality, even though it isn't explicitly imported...

I don't really want to have to install ODC and all it's deps just to test a few lines of code.

jeremyh (approved):
:+1:

Fair enough -- the method is simple enough not to be a big concern!

PR #55, opened 2019-08-08
Minimal versions of some deps

omad:
Old versions of `attrs` and `jsonschema` prevent eodatasets3 from even being imported.

Pip doesn't always check versions, but it's getting better and it's good to document these requirements.


jeremyh (approved):
Great, thanks

PR #66, opened 2019-12-16
handle empty arrays

alexgleith:
Overview image writing will fail if there is all nodata in a band. This fixes it.

### DIFF
--- a/eodatasets3/images.py
+++ b/eodatasets3/images.py
@@ -784,7 +784,7 @@ def read_valid_mask_and_value_range(
         if calculate_percentiles is not None:
             the_data = array[valid_data_mask]
             # Check if there's a non-empty array first
-            if the_data:
+            if the_data.any():
                 low, high = np.percentile(
                     the_data, calculate_percentiles, interpolation="nearest"
                 )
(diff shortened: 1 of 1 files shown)


jeremyh:
Thanks Alex! It looks like it just needs a `black` format to pass the checks.

jeremyh:
Install pre-commit to make it automatic: https://github.com/GeoscienceAustralia/eo-datasets#pre-commit-setup

jeremyh:
... hm, I'm guessing you did, but a new version of Black changed the formatting slightly. We're pinned to `stable`, not a specific version, but it's not exactly stable yet....

jeremyh:
Black issue addressed in #67

jeremyh:
Now it's a test failure
```
>   # Check if there's a non-empty array first
>   if the_data:
E   ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

alexgleith:
I think the tests pass now @jeremyh :-)

jeremyh (approved):
:+1: Thanks Alex

PR #68, opened 2020-01-15
Add a method to add a uuid directory to the output file structure.

dunkgray:
A change to eodatasets to get the uuid directory functionality.  This is confined to a method and the method will not be called in the eodatasets code.   I'm hoping this can be an interum soln to use until we get the full go-ahead.

### DIFF
--- a/eodatasets3/assemble.py
+++ b/eodatasets3/assemble.py
@@ -632,19 +632,21 @@ class DatasetAssembler(EoFields):
 
         self._software_versions.append(dict(name=name, url=url, version=version))
 
-    def add_uuid_to_output_location(self):
+    def add_uuid_to_output_location(self, last_dir=None):
         """
-        Add the dataset_id to the output location.
-        Warning: It's assumed this function will only be used for sentinel 2 data.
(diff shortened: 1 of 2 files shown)


dunkgray:
hey Jeremy, I've finished adding more code.  Fell free to review now.

jeremyh (changes requested):
Thanks Duncan! Nice work.

As discussed, we might change the public api to be consistent with the other public options. Rather than a separate function, it can be a new option for the `naming_conventions` parameter.

eg.
```python
 with DatasetAssembler(
            Path('/some/output/collection/path'), 
            naming_conventions='dea-uuid') as p:
    ...
```

And perhaps change the s2 test/s to use it? https://github.com/GeoscienceAustralia/eo-datasets/blob/eodatasets3/tests/integration/test_assemble.py#L243

alexgleith:
Hey @dunkgray we walked this back, right? I guess we can leave in the UUID option, but we don't actually want to do it for Sentinel-2, was my understanding.

dunkgray:
I've put in a new pull request

PR #73, opened 2020-03-17
The wagl Gadi migration should bump the second version digit, not the third

jeremyh:
Each metadata version number has three digits:  eg: "3.0.0".

The Previous PR bumped the third digit (ie. "3.0.1") for new gadi data. But the third digit is the version of each dataset, not the collection. It's intended to be bumped when a dataset is reprocessed.

This changes the bump for our gadi software change to be the second digit, ie: "3.1.0"

I'm also adding the processing day as the dataset version (ie. "3.1.20190711"), to match what's done for L1 and USGS datasets. Currently wagl packages do no tracking of the dataset version, so reprocesses will all have the same number. This is better default behaviour, and it is backwards compatible with existing datasets.

### DIFF
--- a/tests/integration/test_packagewagl.py
+++ b/tests/integration/test_packagewagl.py
@@ -61,42 +61,42 @@ def test_whole_wagl_package(
     assert_file_structure(
         expected_folder,
         {
-            "ga_ls8c_ard_3-0-1_092084_2016-06-28_final.odc-metadata.yaml": "",
-            "ga_ls8c_ard_3-0-1_092084_2016-06-28_final.proc-info.yaml": "",
-            "ga_ls8c_ard_3-0-1_092084_2016-06-28_final.sha1": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_band01.tif": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_band02.tif": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_band03.tif": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_band04.tif": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_band05.tif": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_band06.tif": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_band07.tif": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_band08.tif": "",
-            "ga_ls8c_nbar_3-0-1_092084_2016-06-28_final_thumbnail.jpg": "",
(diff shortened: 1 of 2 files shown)


sixy6e:
The logic makes sense when the concept of the final number is dataset version.  I missed that part.  I was looking at the collection number as a whole, and thinking this is a patch to the collection, but completely ignoring everything at a dataset level.

sixy6e:
Will be good to confirm this with @simonaoliver

tests/integration/test_packagewagl.py
```diff
@@ -61,42 +61,42 @@ def test_whole_wagl_package(
...
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_nbar-contiguity.tif": "",
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_nbart-contiguity.tif": "",
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_relative-azimuth.tif": "",
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_relative-slope.tif": "",
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_satellite-azimuth.tif": "",
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_satellite-view.tif": "",
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_solar-azimuth.tif": "",
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_solar-zenith.tif": "",
-            "ga_ls8c_oa_3-0-1_092084_2016-06-28_final_time-delta.tif": "",
+            "ga_ls8c_ard_3-1-20190711_092084_2016-06-28_final.odc-metadata.yaml": "",
```
sixy6e, line 64: I'm ok with a date being included as part of the version number `3-1-20190711`, however, that looks like a `yyyymmdd` not a `yyyyddd` for day of year.  Do we need that level though.  Or simply `3.1.0`, and if it gets reprocessed it is `3.1.1`, and if we do an algorithm update, it'll be `3.2.0`.

jeremyh: The tough thing about a numerical dataset version is how to track the reprocesses.

Currently, the processing software works on its own with no knowledge of the outside world.

In order to know that this is exactly the _third_ version of a dataset, it would need to be able to look that up somewhere, which introduces dependencies. A date-based version, like USGS already does with L1, is a much simpler and more reliable solution.

I don't mind YYDOY vs YYMMDD, but I find the latter easier to understand at a glance ("this is from november") -- there's no need for mental conversion, and it's only one character more.

sixy6e: The latter is definitely more understandable, and I'd favour that version.  The DOY might be what USGS use for acquistion date, not processing date.

sixy6e (approved):
I'm good with these changes.  Makes it a lot more clearer when defining a collection version number and dataset version number.

tests/integration/test_packagewagl.py
sixy6e: Go with `YYYYMMDD` as you've already done.

PR #74, opened 2020-04-22
Record scan-gap-interpolation in Landsat L1 metadata

jeremyh:
And bump pre-commit libraries, as the current version of flake8 is failing on some systems.

### DIFF
--- a/eodatasets3/properties.py
+++ b/eodatasets3/properties.py
@@ -180,6 +180,7 @@ _LANDSAT_EXTENDED_PROPS = {
     "landsat:image_quality_oli": None,
     "landsat:image_quality_tirs": None,
     "landsat:processing_software_version": None,
+    "landsat:scan_gap_interpolation": float,
     "landsat:station_id": None,
 }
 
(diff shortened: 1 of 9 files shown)


eodatasets3/properties.py
```diff
@@ -180,6 +180,7 @@ def parsed_sentinel_tile_id(tile_id) -> Tuple[str, Dict]:
     "landsat:image_quality_oli": None,
     "landsat:image_quality_tirs": None,
     "landsat:processing_software_version": None,
+    "landsat:scan_gap_interpolation": float,
```
sixy6e, line 183: Can you confirm if *float* is correct for the scan gap interpolation property?  I thought it'd be a boolean or something.

sixy6e: Is it number of pixels or something?

sixy6e (approved):
Looks good to me.

eodatasets3/properties.py
jeremyh: Yeah, in the lpgs MTLs it's a float. 

In our collections, they're all either '2.0' (usgs) or '0.0' (ours). 

A boolean is probably more in line with how we use the field ourselves today, but it didn't seem worth deviating from the norm.

PR #76, opened 2020-06-11
Updated wagl versioning

jeremyh:
As discussed among @simonaoliver, @sixy6e 

Includes updated tests.

There's also an old commit here from May to remove a `sync` call in the recompressor. `sync` is  *extremely* expensive on gadi and not necessary here.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,7 @@ repos:
       hooks:
           - id: yamllint
     - repo: https://gitlab.com/pycqa/flake8
-      rev: 3.7.9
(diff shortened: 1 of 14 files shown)


jeremyh:
The build failed because of the recent release of flake8 3.8 (which adds new checks). 

... I'd prefer to pin our linters so that we can choose when to make these type of fixes (not on the eve of a bugfix release like this). But the changes are small enough I'll push a fix.

sixy6e (approved):
:+1: 

I'm assuming the converting of the _f-strings_ to normal strings is intended.

PR #77, opened 2020-07-01
Find all image datasets within the HDF5 file, rather than a subset. A…

sixy6e:
…nd skip datasets of type bool as GDAL can't cater for it at this time.

**This is just an example**

### DIFF
--- a/eodatasets3/wagl_md_only.py
+++ b/eodatasets3/wagl_md_only.py
@@ -48,11 +48,13 @@ def package(outdir):
 
 
         with h5py.File(str(INDIR.joinpath(WAGL_FNAME)), 'r') as fid:
-            group = fid[GROUP_PATH]
-            img_paths = [ppjoin(group.name, pth) for pth in find(group, 'IMAGE')]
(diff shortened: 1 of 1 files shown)


eodatasets3/wagl_md_only.py
```diff
@@ -48,12 +48,14 @@ def package(outdir):
...
 
         with h5py.File(str(INDIR.joinpath(WAGL_FNAME)), 'r') as fid:
-            group = fid[GROUP_PATH]
-            img_paths = [ppjoin(group.name, pth) for pth in find(group, 'IMAGE')]
+            img_paths = [ppjoin(fid.name, pth) for pth in find(fid, 'IMAGE')]
 
             for pathname in img_paths:
                 ds = fid[pathname]
 
+                if ds.dtype == 'bool':
```
Oceancolour-RG: consider using numpy.dtype.name

sixy6e: Thanks for tip.  Good pickup!  Explicit type comparison is much safer.

Oceancolour-RG (approved):
Looks good

PR #82, opened 2020-07-10
Lower the gdal cache size

### LINKED ISSUE #81: Excessive memory usage in NCI jobs
On monitored systems such as NCI, the reported memory usage includes any cached objects.  GDAL defaults to using 5% of the available memory.

One suggestion would be to tile the I/O rather than read the whole image at once into memory. This would enable processing of images larger than the available memory, but would require some time to rewrite.  Unfortunately, time is not on our side so this approach will have to wait.

A quicker solution for the short term is to change the _GDAL_CACHEMAX_ environment.  This can be done temporarily at the time of opening any imagery by:

```python
with rasterio.Env(GDAL_CACHEMAX=64):
    with rasterio.open(<filename>) as src:
        # do stuff
```

Where `64` indicates 64MB.

Oceancolour-RG:
Added rasterio.Env(GDAL_CACHEMAX=64) throughout the eo-datasets code

### DIFF
--- a/eodatasets3/assemble.py
+++ b/eodatasets3/assemble.py
@@ -389,15 +389,16 @@ class DatasetAssembler(EoFields):
                         (DEA has measurements called ``blue``, but their written filenames must be ``band04`` by
                         convention.)
         """
-        with rasterio.open(path) as ds:
-            self.write_measurement_rio(
-                name,
-                ds,
-                overviews=overviews,
-                expand_valid_data=expand_valid_data,
-                overview_resampling=overview_resampling,
-                file_id=file_id,
-            )
+        with rasterio.Env(GDAL_CACHEMAX=64):
+            with rasterio.open(path) as ds:
+                self.write_measurement_rio(
+                    name,
+                    ds,
+                    overviews=overviews,
+                    expand_valid_data=expand_valid_data,
+                    overview_resampling=overview_resampling,
+                    file_id=file_id,
+                )
 
     def write_measurement_rio(
         self,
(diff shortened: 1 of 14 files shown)


jeremyh (commented):
Thanks for the submission. It’s useful to put a line in the pull request on why the change is needed.

From my understanding, gdal’s default cache_max is 5% of machine memory, and gadi processing nodes have 24 cores, so a job that runs one instance-per-core, and does nothing else, will be allocating 120% of memory just for gdal_cache. Is that what you’re hitting?

I’m hesitant to hard-code this value as I think the current gdal default is better for most people, and most environments outside of NCI.

I think using gdal’s environment variables in PBS jobs are a better fix: ie. `export GDAL_CACHEMAX=whatever` at the start of your job. People running PBS jobs already need to declare their memory usage elsewhere, so I think it’s better to put the onus on them than the general population. Most NCI jobs I know of don’t run packaging exclusively, so can use a higher value.

jeremyh:
Oh, I missed it: this links to #81. The reasoning is there.

sixy6e:
Ok, so rather generically cater for excessive memory consumption, set the global environment variable 

`GDAL_CACHEMAX`

to something like `64`, for 64Mb.

Later down the line we'll look at a tiling mechanism to handle blocks/tiles of data at a time rather than the entire array.

truth-quark:
> Later down the line we'll look at a tiling mechanism to handle blocks/tiles of data at a time rather than the entire array.

I've done this once or twice in Python before, so there's reuse options. `PyRate` may have this, failing that I know of another repo with similar goods.

Oceancolour-RG:
Thanks for the useful feedback.
I ran the following three tests:
1) original eo-datasets without the `rasterio.Env(GDAL_CACHEMAX=64)`
2) modified eo-datasets with the `rasterio.Env(GDAL_CACHEMAX=64)` hard-coded
3) original eo-datasets without the `rasterio.Env(GDAL_CACHEMAX=64)` this time with `export GDAL_CACHEMAX=64` in the bash script.

I then packaged T147D_F01 using the three different treatments above and unfortunately the memory usage was unchanged, that is, neither setting `rasterio.Env(GDAL_CACHEMAX=64)` nor `export GDAL_CACHEMAX=64` made any difference. I suspect the only way to reduce the amount of memory consumed is to handle blocks/tiles of data at a time.

PR #83, opened 2020-07-13
Support for image "layer" when creating measurement record

dtpc:
Hello,

Currently, using `MeasurementRecord.record_image()` to assemble measurements for an EO3 metadata file does not support hierarchical data formats that use the `"layer"` property.  This PR add this in, trying to touch as little of everything else as possible.

We are looking to use this mostly with `Zarr` formats but it would also be needed for e.g. `NetCDF`.

Cheers,

Dave


jeremyh (approved):
Looks good. Thanks Dave.

PR #84, opened 2020-07-16
Update to stac v1.0.0-beta.1

sachitrajbhandari:
- Added item fields specified in v1.0.0-beta.1 and include absolute links in STAC doc
> - Added stac_version (REQUIRED)
> - Added stac_extensions
> - Added additional field in Asset Object (eo:bands, roles, type)
-  I've taken an approach to include the additional fields in Asset Object  (e.g. type, roles, eo:bands) from the templated JSON file ([ga_ls8c_ard_3.json.txt](https://github.com/GeoscienceAustralia/eo-datasets/files/4951207/ga_ls8c_ard_3.json.txt))
```
"assets": {
  "nbar_nir": {
    "eo:bands": [  {  "name": "nbar_nir"}],
    "type": "image/tiff; application=geotiff",
    "roles": ["data" ]
  }
}
```
- The `'eo3-to-stac'` tool takes two additional optional arguments:
```
 -t '<stac-template.json>' 
 -u '<stac-base-url>'
 -p '<product-base-url>'
```


eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+            #     "rel": "self",
+            #     "href": '?',
+            # },
+            {
+                "rel": "self",
+                "type": "application/json",
+                "href": base_url + output_path.name,
+            },
+            {
+                "title": "STAC's Source Item",
```
alexgleith: I think this should say "Source Dataset YAML"

eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+                "rel": "derived_from",
+                "href": base_url + input_metadata.name,
+            },
+            {"rel": "odc_product", "type": "text/html", "href": dataset.product.href},
```
alexgleith: That link (https://collections.dea.ga.gov.au/product/ga_ls8c_ard_3) isn't actually valid. A valid link is https://explorer.dea.ga.gov.au/product/ga_ls8c_ard_3. 

I'm not sure what it should be. But let's also include a `title` for this.

eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+            {
+                "rel": "alternative",
+                "type": "text/html",
+                "href": f"https://explorer.dea.ga.gov.au/dataset/{dataset.id}",
```
alexgleith: Is this the NCI explorer? Or the production AWS one? I wonder if we want to link to the NCI still?

eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+            **{
+                name: (
+                    {**stac_data["assets"][name], "href": base_url + m.path}
+                    if "assets" in stac_data and name in stac_data["assets"]
+                    else {"href": base_url + m.path}
+                )
+                for name, m in dataset.accessories.items()
+            },
+        },
+        links=[
```
alexgleith, line 225: Do you think we should add a `parent` link to the source document, maybe on THREDDS?

eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+    :param base_url: Base URL path for STAC file
+    :return: Dict of the STAC content
+    """
+
+    project = partial(
+        pyproj.transform, pyproj.Proj(init=dataset.crs), pyproj.Proj(init="epsg:4326"),
+    )
+    wgs84_geometry: BaseGeometry = transform(project, dataset.geometry)
+    item_doc = dict(
+        stac_version=stac_data["stac_version"]
```
alexgleith: If you have an option for an empty dictionary, then this will be hit, and it will break.

eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+        if "stac_version" in stac_data
```
alexgleith: It's better to do 

```
stac_version = stac_data.get("stac_version", "1.0.0-beta.1")
```

eodatasets3/scripts/tostac.py
alexgleith: Actually, I see what you're doing here...

eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+        else "1.0.0-beta.1",
+        stac_extensions=stac_data["stac_extensions"]
+        if "stac_extensions" in stac_data
```
alexgleith: Breaking these over multiple lines is a little confusing and misleading.

eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+        else [],
+        id=dataset.id,
+        type=stac_data["type"] if "type" in stac_data else "Feature",
+        bbox=wgs84_geometry.bounds,
+        geometry=wgs84_geometry.__geo_interface__,
+        properties={**dataset.properties, "odc:product": dataset.product.name},
```
alexgleith: I think this is neat, but are we sure we can do a direct translation of dataset properties to stac standard keys? If there are `eo:gsd` or similar fields, we want to be using the STAC name, not our ODC one.

eodatasets3/scripts/tostac.py
```diff
@@ -19,55 +19,111 @@
...
+    wgs84_geometry: BaseGeometry = transform(project, dataset.geometry)
+    item_doc = dict(
+        stac_version=stac_data["stac_version"]
+        if "stac_version" in stac_data
+        else "1.0.0-beta.1",
+        stac_extensions=stac_data["stac_extensions"]
+        if "stac_extensions" in stac_data
+        else [],
+        id=dataset.id,
+        type=stac_data["type"] if "type" in stac_data else "Feature",
```
alexgleith: Won't they all be a Feature, because we're creating GeoJSON, right?

alexgleith (commented):
I think we need an example stac-template if we're going to use templates.

eodatasets3/scripts/tostac.py
sachitrajbhandari: EO3 field names replaced with STAC standard field names

eodatasets3/scripts/tostac.py
sachitrajbhandari: Fixed

eodatasets3/scripts/tostac.py
sachitrajbhandari: Fixed

eodatasets3/scripts/tostac.py
sachitrajbhandari: Fixed

eodatasets3/scripts/tostac.py
sachitrajbhandari: Added an option to provide `product_base_url` to assign correct link

eodatasets3/scripts/tostac.py
sachitrajbhandari: Tittle added - `Open Data Cube Product`

eodatasets3/scripts/tostac.py
sachitrajbhandari: Fixed

alexgleith (commented):
Can we please include grid handling, as per the Proj extension here: https://github.com/radiantearth/stac-spec/tree/dev/extensions/projection

We want to add the `shape` and `transform` fields

eodatasets3/scripts/tostac.py
sachitrajbhandari: Yes, `Type` is always `Feature`

sachitrajbhandari:
> We want to add the `shape` and `transform` fields

Added `proj` extension with `proj:shape` and `proj:transform` fields

eodatasets3/scripts/tostac.py
```diff
@@ -7,70 +7,182 @@
...
+    :param product_base_url: Base URL for ODC product
+    :return: Dict of the STAC content
+    """
+
+    project = partial(
+        pyproj.transform, pyproj.Proj(init=dataset.crs), pyproj.Proj(init="epsg:4326"),
+    )
+    wgs84_geometry: BaseGeometry = transform(project, dataset.geometry)
+    item_doc = dict(
+        stac_version=stac_data.get("stac_version", "1.0.0-beta.1"),
```
jeremyh: I'm not sure about us taking stac version from both sources. 

This script adds properties for a specific version of stac, and if someone else's template uses a different version of stac there will be an inconsistent result. 

I think I'd prefer just to throw an error if their template doesn't use the same version of stac as this script. (... and some time in the future, if needed, we could add multiple stac version support)

eodatasets3/scripts/tostac.py
```diff
@@ -7,70 +7,182 @@
...
 
-        echo(output_path)
+
+def create_stac(
+    dataset, input_metadata, output_path, stac_data, stac_base_url, product_base_url
+):
+    """
+    Create STAC document
+
+    :param dataset: Dict of the metadata content
```
jeremyh: It's not a high priority, but in most of this repo we're using type annotations in the function parameters rather than describing the type in English (see the `run()` function above).

eodatasets3/scripts/tostac.py
```diff
@@ -7,70 +7,182 @@
...
+        return [value]
+    else:
+        return value
+
+
 @click.command(help=__doc__)
+@click.option("--stac-template", "-t", type=str)
+@click.option("--stac-base-url", "-u", default="", type=str)
+@click.option(
+    "--product-base-url", "-p", default="https://explorer.dea.ga.gov.au", type=str
```
jeremyh: The code below assumes that this URL is an Explorer address (it's computing URL offsets from it that are specific to Explorer), so we probably should name the parameter more specifically: "explorer-url" or something

eodatasets3/scripts/tostac.py
```diff
@@ -7,70 +7,182 @@
...
+    :return: Dict of the STAC content
+    """
+
+    project = partial(
+        pyproj.transform, pyproj.Proj(init=dataset.crs), pyproj.Proj(init="epsg:4326"),
+    )
+    wgs84_geometry: BaseGeometry = transform(project, dataset.geometry)
+    item_doc = dict(
+        stac_version=stac_data.get("stac_version", "1.0.0-beta.1"),
+        stac_extensions=stac_data.get("stac_extensions", []),
```
jeremyh: This should add the stac extensions we use.. proj? Maybe 'eo', though it doesn't have all 'eo' required fields.

jeremyh (commented):
This looks good. The script is real enough now that we might want to start looking at some tests like the other scripts in this repo. (Luckily it's a pretty easy script to test: an input document produces an expected output document, and stac has schemas for validation)

eodatasets3/scripts/tostac.py
```diff
@@ -7,70 +7,182 @@
...
+        type="Feature",
+        bbox=wgs84_geometry.bounds,
+        geometry=wgs84_geometry.__geo_interface__,
+        properties={
+            **{
+                MAPPING_EO3_TO_STAC.get(key, key): convert_value_to_stac_type(key, val)
+                for key, val in dataset.properties.items()
+            },
+            "odc:product": dataset.product.name,
+            "proj:epsg": dataset.crs.lstrip("epsg:"),
```
jeremyh: This might fail badly for non-epsg definitions. In the short term we can just throw a non-implemented exception for those (I prefer that to silently corrupt results)

eodatasets3/scripts/tostac.py
```diff
@@ -7,70 +7,182 @@
...
+            {
+                "title": "Source Dataset YAML",
+                "rel": "derived_from",
+                "href": urljoin(stac_base_url, input_metadata.name),
+            },
+            {
+                "title": "Open Data Cube Product",
+                "rel": "odc_product",
+                "type": "text/html",
+                "href": urljoin(product_base_url, f"product/{dataset.product.name}"),
```
jeremyh: This is now linking to a human-readable html view of the product. I think I'd prefer to keep the "odc_product" rel name for links to the actual machine-readable (yaml/json) product. 

Maybe name it `product_overview` or something else... I'm not sure what the norm is with made-up `rel`s.

eodatasets3/scripts/tostac.py
alexgleith: I agree. Just including the numbers here is enough, splitting on `:` is hopefully safe...

It's a problem with the `proj` extension though, as assuming `epsg` as the prefix is not right.

jeremyh: It looks like in the spec the field is mandatory, but the value can be null, in which case one of the others like `proj:wkt2` is specified for non-epsg.

eodatasets3/scripts/tostac.py
sachitrajbhandari: `odc_product` changed to `product_overview`

eodatasets3/scripts/tostac.py
sachitrajbhandari: `proj:epsg` value safely converted to `int` else set `None`.

eodatasets3/scripts/tostac.py
sachitrajbhandari: Added used stac extension

eodatasets3/scripts/tostac.py
sachitrajbhandari: Changed to `explorer-base-url`

eodatasets3/scripts/tostac.py
sachitrajbhandari: Used type annotations for consistent code style

eodatasets3/scripts/tostac.py
sachitrajbhandari: Added STAC version check

sachitrajbhandari:
> This looks good. The script is real enough now that we might want to start looking at some tests like the other scripts in this repo. (Luckily it's a pretty easy script to test: an input document produces an expected output document, and stac has schemas for validation)

Added STAC validation and Test module.

tests/integration/test_tostac.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_tostac(input_doc_folder: Path):
+    input_metadata_path = input_doc_folder.joinpath(ODC_METADATA_FILE)
+    assert input_metadata_path.exists()
+
+    input_template_path = input_doc_folder.joinpath(STAC_TEMPLATE_FILE)
+    assert input_template_path.exists()
+
+    run_prepare_cli(
```
alexgleith: Why are we writing files here? Wouldn't it be better doing this in memory?

eodatasets3/scripts/tostac.py
```diff
@@ -82,5 +228,26 @@ def json_fallback(o):
     )
 
 
+def validate_stac(item_doc: Dict):
+    # Validates STAC content against schema of STAC item and STAC extensions
+    stac_content = json.loads(json.dumps(item_doc, indent=4, default=json_fallback))
```
alexgleith: Loads and dumps, why?

eodatasets3/scripts/tostac.py
```diff
@@ -82,5 +228,26 @@ def json_fallback(o):
...
+        schema_urls.append(
+            f"https://schemas.stacspec.org/"
+            f"v{item_doc.get('stac_version')}"
+            f"/extensions/{extension}"
+            f"/json-schema/schema.json#"
+        )
+
+    for schema_url in schema_urls:
+        schema_json = requests.get(schema_url).json()
+        validate(stac_content, schema_json)
```
alexgleith, line 282: This is neat!

eodatasets3/scripts/tostac.py
```diff
@@ -5,72 +5,218 @@
...
+    "--explorer-base-url",
+    "-e",
+    default="https://explorer.dea.ga.gov.au",
+    type=str,
+    help="Base URL of the ODC Explorer",
+)
+@click.option(
+    "--validate/--no-validate",
+    default=True,
+    help="Flag it for stac document validation. By default flagged",
```
alexgleith: What's the logic for validating here?

eodatasets3/scripts/tostac.py
```diff
@@ -5,72 +5,218 @@
...
+            explorer_base_url,
+            validate,
+        )
+
         with output_path.open("w") as f:
             json.dump(item_doc, f, indent=4, default=json_fallback)
 
-        echo(output_path)
+
+def create_stac(
```
alexgleith: Maybe this should be dc_to_stac?

eodatasets3/scripts/tostac.py
```diff
@@ -5,72 +5,218 @@
...
+    do_validate: bool,
+) -> dict:
+    """
+    Creates a STAC document
+    """
+
+    stac_ext = ["eo", "view", "projection"]
+    stac_ext.extend(stac_data.get("stac_extensions", []))
+
+    project = partial(
```
alexgleith: We should be consistent in how we do this. Using the ODC function is what we have done a few times in the past: https://github.com/opendatacube/odc-tools/blob/develop/libs/index/odc/index/stac.py#L72

sachitrajbhandari: Used ODC function for CRS transform

eodatasets3/scripts/tostac.py
sachitrajbhandari: Done.

eodatasets3/scripts/tostac.py
sachitrajbhandari: Providing an option to validate or not during stac conversion

tests/integration/test_tostac.py
sachitrajbhandari: To have complete test of `eo3-to-stac` which takes metadata files as input.

eodatasets3/scripts/tostac.py
sachitrajbhandari: Thanks!

eodatasets3/scripts/tostac.py
sachitrajbhandari: Using dumps to JSON serialisation of objects like `UUID` to string and loads to convert back to Python objects

eodatasets3/scripts/tostac.py
```diff
@@ -2,75 +2,214 @@
...
+    )
+    if do_validate:
+        validate_stac(item_doc)
+    return item_doc
 
 
 def json_fallback(o):
     if isinstance(o, datetime):
-        return o.isoformat()
+        return f"{o.isoformat()}Z"
```
jeremyh: Only add a `Z` suffix if `o.tzinfo` is null, because otherwise timezone may be doubled-up.

eodatasets3/scripts/tostac.py
```diff
@@ -2,75 +2,214 @@
...
+    "--explorer-base-url",
+    "-e",
+    default="https://explorer.dea.ga.gov.au",
+    type=str,
+    help="Base URL of the ODC Explorer",
+)
+@click.option(
+    "--validate/--no-validate",
+    default=True,
+    help="Flag it for stac document validation. By default flagged",
```
jeremyh: For consistent help text, maybe change this to something like "Validate output Stac Item against online schemas". There's no need to mention that it's a flag.

And I'd default it to False because it hits the network and will be slow.

eodatasets3/scripts/tostac.py
```diff
@@ -2,75 +2,214 @@
...
+            "proj:transform": dataset.grids["default"].transform,
+        },
+        # TODO: Currently assuming no name collisions.
+        assets={
+            **{
+                name: (
+                    {
+                        **stac_data.get("assets", {}).get(name, {}),
+                        "href": urljoin(stac_base_url, m.path),
+                        "proj.shape": dataset.grids[
```
jeremyh: Should this be a colon? `proj:shape`?

That's what I see in the example: https://github.com/radiantearth/stac-spec/blob/4988a40f6071666d345e4a1c5adc9a6338973769/extensions/projection/examples/example-landsat8.json#L69

The Stac schemas sadly wont catch errors like this, since they allow any unknown properties.

eodatasets3/scripts/tostac.py
```diff
@@ -2,75 +2,214 @@
...
+                            m.grid if m.grid else "default"
+                        ].shape,
+                        "proj.transform": dataset.grids[
```
jeremyh: Similar as above: `proj:transform`?

setup.py
```diff
@@ -66,14 +66,17 @@
...
         "pyproj",
         "rasterio",
         "ruamel.yaml",
         "shapely",
         "structlog",
         "xarray",
+        "requests",
+        "datacube",
     ],
+    dependency_links=["https://packages.dea.ga.gov.au"],
```
jeremyh: Is this added because we need a beta version of datacube or something? 

We should be able to just use default pypi version, I think?

tests/integration/test_tostac.py
```diff
@@ -0,0 +1,61 @@
...
+        tostac.run,
+        "-t",
+        input_template_path.as_posix(),
+        "-u",
+        "http://dea-public-data-dev.s3-ap-southeast-2.amazonaws.com/"
+        "analysis-ready-data/ga_ls8c_ard_3/088/080/2020/05/25/",
+        "-e",
+        "https://explorer.dev.dea.ga.gov.au/",
+        "--validate",
```
jeremyh: I'm not a big fan of us depending on the internet and downloading the validation schemas during every test run  (and possibly repeating it for every individual test). 

I wonder if there's an easy way for the tool to cache them locally the first time it's run?

tests/integration/test_tostac.py
```diff
@@ -0,0 +1,61 @@
...
+
+TO_STAC_DATA: Path = Path(__file__).parent.joinpath("data/tostac")
+ODC_METADATA_FILE: str = "ga_ls8c_ard_3-1-0_088080_2020-05-25_final.odc-metadata.yaml"
+STAC_TEMPLATE_FILE: str = "ga_ls_ard_3_stac_item.json"
+STAC_EXPECTED_FILE: str = "ga_ls8c_ard_3-1-0_088080_2020-05-25_final.stac-item_expected.json"
+
+deep_diff = partial(DeepDiff, significant_digits=6)
+
+
+def test_tostac(input_doc_folder: Path):
```
jeremyh, line 21: Very nice test :)

This one is currently trusting that the app validates when we ask it to. Maybe add a second test that deliberately adds an invalid property before calling it? (optional)

tests/integration/data/tostac/ga_ls_ard_3_stac_item.json
```diff
@@ -0,0 +1,173 @@
...
+            "type": "image/jpeg",
+            "roles": ["thumbnail"]
+        },
+        "thumbnail:nbart": {
+            "title": "Thumbnail image",
+            "type": "image/jpeg",
+            "roles": ["thumbnail"]
+        },
+        "checksum:sha1": {"type": "text/plain"},
+        "metadata:processor": {"type": "text/plain", "roles": ["metadata"]}
```
jeremyh: The checksum/metadata fields are common and expected. Maybe the tool itself should fill in the type for assets, rather than the user specifying it. 

Using their file extensions,  I'd expect checksum to be text and the metadata here to be yaml.

eodatasets3/scripts/tostac.py
jeremyh: I do the same in some tests for normalisation before comparing things. The indent could be removed though.

Edit: this was a reply to a previous comment that was detached.

eodatasets3/scripts/tostac.py
```diff
@@ -2,75 +2,214 @@
...
+        return [value]
+    else:
+        return value
 
 
 @click.command(help=__doc__)
+@click.option(
+    "--stac-template",
+    "-t",
+    type=str,
```
jeremyh: Use type `PathPath`, similar to the other file path options.

eodatasets3/scripts/tostac.py
```diff
@@ -2,75 +2,214 @@
...
+        bbox=wgs84_geometry.boundingbox,
+        geometry=wgs84_geometry.json,
+        properties={
+            **{
+                MAPPING_EO3_TO_STAC.get(key, key): convert_value_to_stac_type(key, val)
+                for key, val in dataset.properties.items()
+            },
+            "odc:product": dataset.product.name,
+            "proj:epsg": int(dataset.crs.lstrip("epsg:")) if dataset.crs else None,
+            "proj:shape": dataset.grids["default"].shape,
```
jeremyh: A 'default' grid is not guaranteed to be there... it's just the name used when they don't specify a grid name. The least-surprising behaviour would be to pick the most frequently-used grid (since I usually see all bands bar-one use the same grid)

(though I haven't looked into how people usually treat datasets with multple grids with this extension)

eodatasets3/scripts/tostac.py
```diff
@@ -2,75 +2,214 @@
...
+    type=str,
+    help="File path to the STAC document template with static content",
+)
+@click.option(
+    "--stac-base-url", "-u", default="", type=str, help="Base URL of the STAC file"
+)
+@click.option(
+    "--explorer-base-url",
+    "-e",
+    default="https://explorer.dea.ga.gov.au",
```
jeremyh: I know the code previously did this, though I wonder if we should have no default here. This tool may be used by people outside GA. And even within GA we have multiple Explorer instances unfortunately.

(best case: it would be neat to somehow auto-detect it)

jeremyh (changes requested):
This is shaping-up well -- nice work @sachitrajbhandari!

The biggest thing I'm hesitant about is the (effectively mandatory) use of the templates to use this tool now.

They look very cumbersome and error-prone to write, requiring careful Stac knowledge, and without them the Stac Item wont be valid (it's missing required `eo:` etc fields). 

This kind of effort and knowledge is what I'd like this tool to save the user from.

I think taking an odc product definition would be a nicer interface than taking a template, as users already need to write product definitions for their datasets. And product definitions already contain all the data that's in the example test template: lists of bands etc.

The `eo3-validate` command in this repo accepts a product definition as an argument already, so accepting one here might be nice symmetry with that tool.

tests/integration/test_tostac.py
jeremyh: (But this is low-priority and I'm expanding the task list beyond immediate needs. Validate was a nice addition. Alternatively, short term, just don't run validate within these tests.)

eodatasets3/scripts/tostac.py
jeremyh: Auto-detection crazy-idea: Load `https://explorer.dea.ga.gov.au/products.txt` [among other known hosts] and see if product name is there. Though, like the stac validation, it probably should be cached somehow so we're not hitting it all the time

jeremyh: ... though that wouldn't work for brand-new products.

If we were being _that_ crazy, we could technically download the product definition too to get band/other information.

eodatasets3/scripts/tostac.py
sachitrajbhandari: Check for tzinfo added

eodatasets3/scripts/tostac.py
sachitrajbhandari: Changes made as suggested.

eodatasets3/scripts/tostac.py
sachitrajbhandari: Fixed

eodatasets3/scripts/tostac.py
sachitrajbhandari: Fixed

setup.py
sachitrajbhandari: Dependency links removed

tests/integration/data/tostac/ga_ls_ard_3_stac_item.json
sachitrajbhandari: Mime type is added based on their file extensions. What would be the mimetype for metadata yaml file. Any suggestions?

eodatasets3/scripts/tostac.py
sachitrajbhandari: Removed `--stac-template` options

eodatasets3/scripts/tostac.py
sachitrajbhandari: Check added if no `default` grid, fields `proj:shape` and `proj:transform` are not added.

eodatasets3/scripts/tostac.py
sachitrajbhandari: For now, `default` value for `--explorer-base-url` is removed and made optional. Auto-detection to be implemented.

tests/integration/data/tostac/ga_ls_ard_3_stac_item.json
jeremyh: There's contentious answers to that as there seems to be standard ones no one is using. Explorer is currently using 'text/yaml'.

sachitrajbhandari:
> This is shaping-up well -- nice work @sachitrajbhandari!
> 
> The biggest thing I'm hesitant about is the (effectively mandatory) use of the templates to use this tool now.
> 
> They look very cumbersome and error-prone to write, requiring careful Stac knowledge, and without them the Stac Item wont be valid (it's missing required `eo:` etc fields).
> 
> This kind of effort and knowledge is what I'd like this tool to save the user from.
 
Thanks for your thorough review. I'm removed the use of the templates and implemented a dynamic way to fill in the fields for `assets` objects previously taken from templates. I didn't use product definition as needed information could be retrieved from metadata yaml file. Looking forward for your feedback on this approach.

sachitrajbhandari:
@jeremyh More fixes - Changed `mime type` for yaml file and custom `rel` for source yaml link. Looking for your feedback.

jeremyh (approved):
:+1:  Looks good to me. The lack of template is great -- it now "just works" if you run it with a metadata file.

Nice work @sachitrajbhandari

PR #86, opened 2020-07-17
Updates to include clearer names and bool type datasets

sixy6e:
Updating to use oa as the product group instead of parent names for items such as the solar geometry. This removes the duplication in the measurement name and brings it into line with the idea ard C3 terminology.

Also enables boolean datasets. As GDAL doesn't support compound types, we simply convert the dataset to a GeoTIFF of type uint8.


sixy6e:
Will close and establish a clean branch. This one still has binaries in the commit history and lots of conflicts arising from commits mid-way through the commit history.

PR #92, opened 2020-08-10
Ard metadata only

ohehir:


sixy6e:
Something has gone wrong in your code update. You've got two modules:

* eodatasets
* eodatasets3

PR #93, opened 2020-08-10
eodatasets dir removed

ohehir:
@sixy6e

### DIFF
--- a/eodatasets3/tesp_ard_metadata_only.py
+++ b/eodatasets3/tesp_ard_metadata_only.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#/usr/bin/env python
 
 from posixpath import join as ppjoin
 from pathlib import Path
@@ -11,7 +11,7 @@ from eodatasets3 import DatasetAssembler, images, utils
 import eodatasets3.wagl
 from eodatasets3.serialise import loads_yaml
 from boltons.iterutils import get_path
-
+import numpy as np
 
 GDAL_H5_FMT = 'HDF5:"{filename}":/{dataset_pathname}'
 
@@ -30,6 +30,9 @@ def package_non_standard(outdir, granule):
 
     outdir = Path(outdir)
     out_fname = Path(str(granule.wagl_hdf5).replace("wagl.h5", "yaml"))
+    out_fname_bool = Path(str(granule.wagl_hdf5).replace("wagl.h5", "bool.wagl.h5"))
+
+    f = h5py.File(outdir.joinpath(out_fname_bool))
 
     with DatasetAssembler(metadata_path=out_fname, naming_conventions="dea") as da:
         level1 = granule.source_level1_metadata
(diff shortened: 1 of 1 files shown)


eodatasets3/tesp_ard_metadata_only.py
```diff
@@ -30,6 +30,9 @@ def package_non_standard(outdir, granule):
 
     outdir = Path(outdir)
     out_fname = Path(str(granule.wagl_hdf5).replace("wagl.h5", "yaml"))
+    out_fname_bool = Path(str(granule.wagl_hdf5).replace("wagl.h5", "bool.wagl.h5"))
```
sixy6e, line 33: Instead of "bool.wagl.h5" use something that provides a little more context such as:

"converted-datasets.h5"

eodatasets3/tesp_ard_metadata_only.py
```diff
@@ -129,6 +134,34 @@ def package_non_standard(outdir, granule):
...
+                    )
+
+                    # Create output dataset
+                    out_ds = f.create_dataset(measurement_name, ds.shape, dtype=np.uint8)
+                    # get numpy array from input dataset
+                    np_ds = ds.value
+                    # convert boolean array to unit8
+                    np_ds_int = np.uint8(np_ds[:])
+                    # set nodata values from 0 to 255
+                    np_ds_int[np_ds_int == 0] = 255
```
sixy6e, line 150: Don't convert the zeros to 255. leave as is.

eodatasets3/tesp_ard_metadata_only.py
```diff
@@ -129,6 +134,34 @@ def package_non_standard(outdir, granule):
...
+                    # write numpy array back to output h5
+                    out_ds = np_ds_int
```
sixy6e, line 152: You changing the value of `out_ds`

It should be

```python
out_ds[:] = np_ds_int
```

eodatasets3/tesp_ard_metadata_only.py
```diff
@@ -129,6 +134,34 @@ def package_non_standard(outdir, granule):
...
                 # if we are of type bool, we'll have to convert just for GDAL
                 if ds.dtype.name == "bool":
+
+                    pathname = GDAL_H5_FMT.format(
+                        filename=str(outdir.joinpath(out_fname_bool)),
+                        dataset_pathname=pathname,
+                    )
+
+                    # Create output dataset
+                    out_ds = f.create_dataset(measurement_name, ds.shape, dtype=np.uint8)
```
sixy6e, line 144: In one line we can do something like this:

```python
out_ds = f.create_dataset(measurement_name, data=np.uint8(ds[:]), compression='lzf', shuffle=True, chunks=ds.chunks)
```

For the attrs:

```python
for k, v in ds.attrs.items():
    out_ds.attrs[k] = v
```

sixy6e (changes requested):
A few things to fix

sixy6e:
For Fmask, a simple thing such as:

```python
with rasterio.open(<filename>) as ds:
    out_ds = fid.create_dataset(<dataset-name>, data=ds.read(1), compression='lzf', shuffle=True)
    out_ds.attrs['crs_wkt'] = ds.crs.wkt
    out_ds.attrs['geotransform'] = ds.transform.to_gdal()
    out_ds.attrs['description'] = "Converted from ERDAS Imagine format to HDF5 to work with the limitations of varied formats within ODC"
```

PR #94, opened 2020-08-17
Add filter for valid polygons only in consume_and_get_valid_data()

mpaget:
We found some cases where
```
[
    shapely.geometry.shape(shape)
    for shape, val in rasterio.features.shapes(mask)
    if val == 1
]
```
creates invalid polygons. We haven't investigated why invalid polygons are created. Example scene is LE07_L1TP_124050_20170420_20170516_01_T1 processed to SR via ESPA.

This PR just introduces a filter to remove invalid polygons before passing to shapely.ops.unary_union().

### DIFF
--- a/eodatasets3/images.py
+++ b/eodatasets3/images.py
@@ -310,17 +310,23 @@ class MeasurementRecord:
 
         (they are consumed in order to to minimise peak memory usage)
         """
+
+        def valid_shape(shape):
+            if shape.is_valid:
+                return shape
+            return shape.buffer(0)
+
         geoms = []
         while self.mask_by_grid:
             grid, mask = self.mask_by_grid.popitem()
             mask = mask.astype("uint8")
-            shapes = [
-                shapely.geometry.shape(shape)
-                for shape, val in rasterio.features.shapes(mask)
-                if val == 1
-            ]
-            shapes = [shape for shape in shapes if shape.is_valid]
-            shape = shapely.ops.unary_union(shapes)
+            shape = shapely.ops.unary_union(
+                [
+                    valid_shape(shapely.geometry.shape(shape))
+                    for shape, val in rasterio.features.shapes(mask)
+                    if val == 1
+                ]
+            )
             shape_y, shape_x = mask.shape
             del mask
 
--- a/eodatasets3/images.py
+++ b/eodatasets3/images.py
@@ -315,9 +315,9 @@ class MeasurementRecord:
             grid, mask = self.mask_by_grid.popitem()
             mask = mask.astype("uint8")
             shapes = [
-                    shapely.geometry.shape(shape)
-                    for shape, val in rasterio.features.shapes(mask)
-                    if val == 1
+                shapely.geometry.shape(shape)
+                for shape, val in rasterio.features.shapes(mask)
+                if val == 1
             ]
             shapes = [shape for shape in shapes if shape.is_valid]
             shape = shapely.ops.unary_union(shapes)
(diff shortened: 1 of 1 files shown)


mpaget:
Struggling to replicate the LE07_L1TP_104078_20130429_20161124_01_T1.tar error raised by Travis.

Can you check in your environment, please?

With this PR:
```
python eodatasets3/prepare/landsat_l1_prepare.py --output-base /opt/working/tmp /opt/eo-datasets/tests/integration/data/LE07_L1TP_104078_20130429_20161124_01_T1.tar
2020-08-17 04:28:42,832 INFO Processing /opt/eo-datasets/tests/integration/data/LE07_L1TP_104078_20130429_20161124_01_T1.tar
2020-08-17 04:28:42,866 INFO Ring Self-intersection at or near point 9 50
2020-08-17 04:28:42,881 INFO Ring Self-intersection at or near point 9 50
2020-08-17 04:28:42,899 INFO Checksumming PosixPath('/opt/working/tmp/104/078/LE07_L1TP_104078_20130429_20161124_01_T1.odc-metadata.yaml')
2020-08-17 04:28:42,906 INFO Wrote dataset f23c5fa2-3321-5be9-9872-2be73fee12a6 to /opt/working/tmp/104/078/LE07_L1TP_104078_20130429_20161124_01_T1.odc-metadata.yaml
```

Without this PR:
```
python eodatasets3/prepare/landsat_l1_prepare.py --output-base /opt/working/tmp /opt/eo-datasets/tests/integration/data/LE07_L1TP_104078_20130429_20161124_01_T1.tar
2020-08-17 04:29:12,079 INFO Processing /opt/eo-datasets/tests/integration/data/LE07_L1TP_104078_20130429_20161124_01_T1.tar
2020-08-17 04:29:12,117 ERROR TopologyException: Input geom 1 is invalid: Ring Self-intersection at or near point 9 50 at 9 50
...
    valid_data = self._measurements.consume_and_get_valid_data()
  File "/env/lib/python3.6/site-packages/eodatasets3/images.py", line 320, in consume_and_get_valid_data
    for shape, val in rasterio.features.shapes(mask)
  File "/env/lib/python3.6/site-packages/shapely/ops.py", line 152, in unary_union
    return geom_factory(lgeos.methods['unary_union'](collection))
  File "/env/lib/python3.6/site-packages/shapely/geometry/base.py", line 76, in geom_factory
    raise ValueError("No Shapely geometry can be created from null value")
```

jeremyh:
That's interesting. With the patch, that test has a slightly different boundary:

Screenshot_2020-08-19 geojson io
(red is with patch, blue without)

From running in debug, it looks like the test data includes an invalid shape, but comes out ok. I'm not yet sure which is the correct behaviour...

jeremyh:
Without looking into it further,  my gut feeling is that dropping some shapes may be a bad idea in some circumstances, as it could break our promise to datacube that the valid_data shape contains all of the valid pixels (it's allowed to be bigger than the true valid pixel area, but not smaller, I believe).

We might want to try one of the methods to fix invalid shapes rather than dropping them completely.

jeremyh:
Ok, the tests passes if I `shape.buffer(0)` any invalid shapes (they're now all valid and so retained). I wonder if that will fix your original issue, so we wont need to drop them at all?

```python
def valid_shape(shape):
    if shape.is_valid:
        return shape
    return shape.buffer(0)

shapes = [
    valid_shape(shapely.geometry.shape(shape))
    for shape, val in rasterio.features.shapes(mask)
    if val == 1
]
```

mpaget:
Super. Thank you for looking into this. Yes, I suspect your buffer proposal will work nicely. I'll try it at our end, and if all well will update this PR.

jeremyh (approved):
Looks good -- thanks Matt.

PR #97, opened 2020-09-30
C3 version fix

n3h3m:
- Adds collection number new field
- The destination filepath follows a new format ex: `ga_ls_wo_3/1-6-0/090/081/1998/07/30/ga_ls_wo_3_090081_1998-07-30_interim.odc-metadata.yaml`


jeremyh:
Thanks for looking into this Nehem.

Is this the new standard it's implementing? https://docs.google.com/spreadsheets/d/10CawHQvvwiyPojBRezPsYIQkLVtMtJL4xbpqrxJIpcc/edit#gid=1327870591

As mentioned on Slack, we probably need to maintain backwards compatibility with the older folder structure, as many products are already produced and stored using the older structure. The Travis tests will fail as they're testing for this consistency.

We will probably need to make this a new option under the `DatasetAssembler(naming_conventions="...")` property. Currently users specify `naming_conventions="dea"`, so this could be `"dea-2020"` or something.

jeremyh (commented):
(after Slack discussion) -- re-request review when ready.

eodatasets3/assemble.py
```diff
@@ -169,6 +170,8 @@ def __init__(
             self.names = ComplicatedNamingConventions.for_standard_dea(self)
         elif naming_conventions == "dea_s2":
             self.names = ComplicatedNamingConventions.for_standard_dea_s2(self)
+        elif naming_conventions == "dea_c3":
+            self.names = ComplicatedNamingConventionsAlchemist.for_c3_processing(self)
```
alexgleith: Maybe we should name this `for_c3_derivatives`?

eodatasets3/model.py
```diff
@@ -373,6 +372,81 @@ def producer_abbreviated(self) -> Optional[str]:
             )
 
 
+class ComplicatedNamingConventionsAlchemist(ComplicatedNamingConventions):
+
+    """
+    This class is inherited from ComplicatedNamingConventions
+    and overrides few attributes specific to C3 dataprocessing at the Alchemist end.
```
alexgleith:
```suggestion
    and overrides few attributes specific to C3 data processing for data other than ARD.
```

eodatasets3/model.py
```diff
@@ -373,6 +372,81 @@ def producer_abbreviated(self) -> Optional[str]:
             )
 
 
+class ComplicatedNamingConventionsAlchemist(ComplicatedNamingConventions):
```
alexgleith: Again, I think `derivatives` rather than `alchemist` perhaps...

eodatasets3/model.py
```diff
@@ -373,6 +372,81 @@ def producer_abbreviated(self) -> Optional[str]:
...
+            ),
+        )
+
+    @property
+    def _org_collection_number(self) -> Optional[int]:
+
+        # Deliberately fail if collection_number is not defined in the config yaml
+        return int(self.dataset.collection_number)
+
+    def _product_group(self, subname=None) -> str:
```
alexgleith, line 406: These product groups are supposed to have the collection number in them... I'm guessing that comes in somewhere else... but perhaps it shouldn't, and it should come in here?

eodatasets3/model.py
```diff
@@ -373,6 +372,81 @@ def producer_abbreviated(self) -> Optional[str]:
...
+        # Copycat of the parent method except version is simply collection number
+        p = self.dataset
+        version = p.collection_number
+        maturity: str = p.properties.get("dea:dataset_maturity")
+        return "_".join(
+            [
+                p
+                for p in (
+                    self._product_group(sub_name),
+                    version,
```
alexgleith: I don't like this, it's misleading.

The join should just be the `product_group`, which really is `product_id` and then all the other bits. Including version here is not correct.

eodatasets3/properties.py
```diff
@@ -413,10 +415,34 @@ def dataset_version(self) -> str:
...
 
+    @property
+    def collection_number(self) -> str:
+        """
+        The version of the collection.
+        Eg:
+          metadata:
+            product_family: wofs
+            dataset_version: 1.6.0
+            collection_number: 3
```
alexgleith, line 426: This is a weird one, because for our ARD we use the USGS collection number. Not sure what we should actually be using here, should it be `3` (our collection) or `1` (the USGS source collection)?

Perhaps a question for @simonaoliver?

jeremyh: From how I understand the naming system, since it's derived from our collection 3 data it should use that as its collection number. Only the Level 1 processed by USGS is using the USGS collection numbers (I believe?).

In the spec, the first number of the `dataset_version` is the collection number, hence ARD datasets having `dataset_version: 3.0.0` etc -- and that's how the [current collection_number property reads it](https://github.com/GeoscienceAustralia/eo-datasets/blob/c115ee1542f3a4a43f9d076aaec4a3f021692d8b/eodatasets3/model.py#L182-L185). If we want to copy it in a stand-alone property, it should match, I think? (do we need it separate?)

n3h3m: @jeremyh In this case we need separation as the sample agreed upon S3 path as `/ga_ls_wo_3/1-6-0/096/095/2000/10/01/ga_ls_wo_3_096095_2000-10-01_final_water.tiff`

We need `dataset_version` to be `1.6.0` in the path also, `_3`. Alex has different opinion as well, the `_3` part can come from part of product itself.

alexgleith: Check this out, @jeremyh : https://data.dea.ga.gov.au/baseline/ga_ls8c_ard_3/089/076/2016/06/22/ga_ls8c_ard_3-0-0_089076_2016-06-22_final.odc-metadata.yaml

`landsat:collection_number: 1`

jeremyh: > Check this out, @jeremyh : https://data.dea.ga.gov.au/baseline/ga_ls8c_ard_3/089/076/2016/06/22/ga_ls8c_ard_3-0-0_089076_2016-06-22_final.odc-metadata.yaml
> 
> `landsat:collection_number: 1`

Ah, yeah. That's specifically in the landsat namespace (all fields of which match verbatim with USGS's MTL document field names/values), so it's not a normal DEA "collection number" field. It's _always_ usgs collection number

jeremyh: > @jeremyh In this case we need separation as the sample agreed upon S3 path as `/ga_ls_wo_3/1-6-0/096/095/2000/10/01/ga_ls_wo_3_096095_2000-10-01_final_water.tiff`
> 
> We need `dataset_version` to be `1.6.0` in the path also, `_3`. Alex has different opinion as well, the `_3` part can come from part of product itself.

So is the second folder a wofs algorithm version or something? I might have missed these discussions, but in previous discussion everyone was very specific on what each component of `dataset_version` was, and eodatasets had to be fixed to match. Might need @simonaoliver to confirm

jeremyh: For the record, I don't think the standard is perfect, but the only thing worse than having imperfect standard fields, is having imperfect standard fields that are used inconsistently everywhere. So I have to play the character of nitpicky bureaucrat that demands specs are matched :slightly_smiling_face:

eodatasets3/model.py
n3h3m: Spot on 👍 simplified it.

eodatasets3/properties.py
alexgleith: The idea is that collection 3's `_3` is included in the product name (explicitly or implicitly, I don't care here!) and is separate to the product version number.

n3h3m: To sum:
On the Alchemist side 

1 - dataset versions for `wofs`  1.6.0 and for `fc` 2.5.0. 
2 - collection_number is 3
3 - The derivatives s3 filepath structure is `ga_ls_wo_3/1-6-0/096/095/2000/10/01/ga_ls_wo_3_096095_2000-10-01_final_water.tif`

@jeremyh @alexgleith if we are happy, we can merge

eodatasets3/model.py
n3h3m: For the earlier logic, collection_number was simply a part of `dataset_version`. This feature requires two different things from which `collection_number` cannot be parsed from `dataset_version`

eodatasets3/properties.py
alexgleith: I'm happy, so long as 1 and 2 are configurable from Alchemist and not hardcoded! Which I think they are.

alexgleith (dismissed):
👍

eodatasets3/properties.py
sixy6e: To re-ask the same question by @jeremyh 

Is the second folder an algorithm version number?

alexgleith: Sorry @sixy6e 

The second folder is the product version number, which is also called the `odc:dataset_version` in the ARD yamls.

So, look at CMI: https://cmi.ga.gov.au/data-products/dea/119/dea-fractional-cover-landsat, it's the `2.2.1` number.

alexgleith: > In the spec, the first number of the `dataset_version` is the collection number, hence ARD datasets having `dataset_version: 3.0.0` etc -- and that's how the [current collection_number property reads it](https://github.com/GeoscienceAustralia/eo-datasets/blob/c115ee1542f3a4a43f9d076aaec4a3f021692d8b/eodatasets3/model.py#L182-L185). If we want to copy it in a stand-alone property, it should match, I think? (do we need it separate?)

There was a decision in one of Simon's meetings about naming conventions recently where we decoupled product version numbers (what you're calling `dataset_version`) from collection numbers.

So for derived products, we need a new field for the `collection_number`, since it's not the first digit of the product number.

sixy6e: But product version and dataset version are different things. They're not substitutable with one another.

If it is desired to use a product version number, there should be a new field added, i.e. `product_version`.  Not use the product version in the namespace for `dataset_version`.  Otherwise not only are we then inconsistent in file naming, but also in the metadata where a single field would then be used to describe different entities.

The reason I ask is that maybe we should change ARD to be consistent with other products based on the new decision.  In which case the ARD product should use a product version instead of a dataset version.

sixy6e: And looking at the CMI link, the 2.2.1 number is pointing to Collection-2. The first digit `2` there is referring to the collection.

alexgleith: Regarding the first digit of a `product_version` being the collection number, we discussed this in Simon's naming conventions meeting, and agreed to not continue this practice. Plus, there are always exceptions, and NIDEM is 1.0.0 currently https://data.dea.ga.gov.au/?prefix=nidem/

But I am curious as to what the difference is between a `dataset_version` and a `product_version`, @sixy6e?

eodatasets3/model.py
```diff
@@ -373,6 +372,75 @@ def producer_abbreviated(self) -> Optional[str]:
...
+            self._displayable_region_code,
+            f"{self.dataset.datetime:%Y-%m-%d}",
+            self.dataset.maturity,
+        ]
+        return "_".join(parts)
+
+    @property
+    def platform_abbreviated(self) -> Optional[str]:
+        # For now from Alchemist the platform is always landsat for C3 processing
+        return "ls"
```
jeremyh: This might cause future accidents (quietly incorrect data). Maybe throw an error if platform doesn't contain "landsat"?

Ideally, long term, we move this logic up to the parent function as a new case.  
 - Presumably? "if platform contains multiple landsats, only show 'ls'"

eodatasets3/properties.py
sixy6e: A `product_version` relates to the tracking of changes relating to the state of the product, which could be algorithmic changes.
A `dataset_version` is a specific instance of a product generated according to the specifications of the product.

Updating the `product_version` most of the time would also update the `dataset_version`.  When it wouldn't is when there are major changes in the definition or generation of the product that deviates significantly from the previous specification.  No longer idempotent, or not within the spec tolerance.  In which a new specification for a product definition, meaning a reset on the `dataset_version`.

The `dataset_version` can be updated independently of the product version and allows for increased traceability and provenance.  For example, updating `eo-datasets`, or other libraries, don't change the product, but changes the way the product was generated to still meet the specifications.  In this instance, the `dataset_version` is updated enabling traceability for a specific instance of a product.

If the intent is to be updating the modules that generate the product, and not record the instance, then we've lost a level of provenance.

Eg generating WOfS with an updated ODC version, may not change the product but is now a different instance of that product.  We can use the same ODC but change the WOfS algorithm (ever so slightly to avoid major differences), is an updated product, as well as an instance of the product.

jeremyh:
To echo previous discussion today: above collection versioning was approved.

From Slack discussion, just fixing: https://github.com/GeoscienceAustralia/eo-datasets/pull/97#discussion_r501385221

At this risk of delaying further (but it’s a quick fix!), I noticed the `collection_number`field is sometimes a string, sometimes an int (in the docstrings). I assume we want it specifically to be an int?

It could be added here to always be normalised and validated as an int: 
https://github.com/GeoscienceAustralia/eo-datasets/blob/334fed42fb91904f039bf71e08dd7e7bffa6979f/eodatasets3/properties.py#L209 
eg. https://github.com/GeoscienceAustralia/eo-datasets/blob/334fed42fb91904f039bf71e08dd7e7bffa6979f/eodatasets3/properties.py#L227

Inconsistency will make using it harder.

eodatasets3/model.py
```diff
@@ -373,6 +372,79 @@ def producer_abbreviated(self) -> Optional[str]:
...
+        ]
+        return "_".join(parts)
+
+    @property
+    def platform_abbreviated(self) -> Optional[str]:
+        # For now from Alchemist the platform is always landsat for C3 processing
+        if "ls" not in self.dataset.platform:
+            raise ValueError(
+                "ComplicatedNamingConventionsDerivatives assumes lansat as platform"
+            )
```
jeremyh: ls -> landsat, to match WOfS and stac etc.

```suggestion
        if "landsat" not in self.dataset.platform:
            raise NotImplementedError(
                f"Only Landsat platforms currently implemented for dea_c3 naming schemes "
                f"(got {self.dataset.platform!r})"
            )
```

jeremyh (approved):
Nice work. 

Good job keeping the git history clean too :slightly_smiling_face:

PR #98, opened 2020-10-01
Further stac conversion improvements, tighten public api

jeremyh:
Expansion of the stac converter so it can be used from other projects.

- Use the stac `created` field instead of ODC's field name.
- Include dataset label as a 'title' property (perhaps there's a better option?)
- Separate the LS8 instruments, following the [spec example](https://github.com/radiantearth/stac-spec/blob/4988a40f6071666d345e4a1c5adc9a6338973769/item-spec/examples/landsat8-sample.json#L48-L51)
- Fall-back to WKT2 when dataset crs is not an epsg code.
- Allow custom URLs, location to be used for the output paths.

- Make all functions private except the main ones. Keep the public api small.
- Add a link to the parent stac collection in Explorer

### DIFF
--- a/eodatasets3/scripts/tostac.py
+++ b/eodatasets3/scripts/tostac.py
@@ -1,5 +1,5 @@
 """
-Convert a new-style ODC metadata doc to a Stac Item.
+Convert an EO3 metadata doc to a Stac Item.
 """
 import json
 from datetime import datetime
@@ -9,14 +9,16 @@ from urllib.parse import urljoin
 from uuid import UUID
 
 import click
+from click import echo, style
 
+import eodatasets3.stac as eo3stac
 from eodatasets3 import serialise
 from eodatasets3.model import DatasetDoc
-import eodatasets3.stac as eo3stac
 from eodatasets3.ui import PathPath
 
 
 @click.command(help=__doc__)
+@click.option("-v", "--verbose", is_flag=True)
 @click.option("--stac-base-url", "-u", help="Base URL of the STAC file")
 @click.option("--explorer-base-url", "-e", help="Base URL of the ODC Explorer")
 @click.option(
@@ -30,6 +32,7 @@ from eodatasets3.ui import PathPath
     nargs=-1,
 )
 def run(
+    verbose: bool,
     odc_metadata_files: Iterable[Path],
     stac_base_url: str,
     explorer_base_url: str,
@@ -48,25 +51,17 @@ def run(
             output_path,
             stac_base_url,
             explorer_base_url,
-            validate,
+            do_validate=False,
         )
+
+        if validate:
+            eo3stac.validate_item(item_doc, log=echo if verbose else lambda line: None)
 
         with output_path.open("w") as f:
             json.dump(item_doc, f, indent=4, default=json_fallback)
 
-
-def json_fallback(o):
-    if isinstance(o, datetime):
-        return f"{o.isoformat()}" if o.tzinfo else f"{o.isoformat()}Z"
-
-    if isinstance(o, UUID):
-        return str(o)
-
-    raise TypeError(
-        f"Unhandled type for json conversion: "
-        f"{o.__class__.__name__!r} "
-        f"(object {o!r})"
-    )
+        if verbose:
+            echo(f'Wrote {style(output_path.as_posix(), "green")}')
 
 
 def dc_to_stac(
@@ -94,9 +89,28 @@ def dc_to_stac(
         explorer_base_url=explorer_base_url,
     )
     if do_validate:
-        eo3stac.validate_stac(doc)
+        eo3stac.validate_item(doc)
 
     return doc
+
+
+def json_fallback(o):
+    """
+    This function is needed here for backwards compatibility.
+
+    Some users try to import it from this script.
+    """
+    if isinstance(o, datetime):
+        return f"{o.isoformat()}" if o.tzinfo else f"{o.isoformat()}Z"
+
+    if isinstance(o, UUID):
+        return str(o)
+
+    raise TypeError(
+        f"Unhandled type for json conversion: "
+        f"{o.__class__.__name__!r} "
+        f"(object {o!r})"
+    )
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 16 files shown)


eodatasets3/scripts/tostac.py
```diff
@@ -163,74 +181,88 @@ def run(
...
+            explorer_base_url=explorer_base_url,
+            do_validate=validate,
         )
 
         with output_path.open("w") as f:
             json.dump(item_doc, f, indent=4, default=json_fallback)
 
 
-def dc_to_stac(
+def dataset_as_stac_item(
```
sachitrajbhandari: This `dc_to_stac` function is used in Landsat C3 sync script and renaming the function name and input parameters (`input_metadata_url, output_url`) will break the code.

eodatasets3/scripts/tostac.py
```diff
@@ -163,74 +181,88 @@ def run(
...
+    if dataset.label:
+        properties["title"] = dataset.label
+
+    # TODO: choose remote if there's multiple locations?
+    dataset_location = dataset.locations[0] if dataset.locations else output_url
+
     item_doc = dict(
         stac_version="1.0.0-beta.2",
-        stac_extensions=["eo", "projection", "view"],
+        stac_extensions=["eo", "projection"],
```
sachitrajbhandari: Extension `view` is used in created STAC doc after EO3 to STAC mapping. e.g.  `"eo:sun_azimuth": "view:sun_azimuth",  "eo:sun_elevation": "view:sun_elevation"`

sachitrajbhandari (commented):
Nice refactoring with limited public functions. Does changes in `created` field and `instruments` values will required updating existing STAC doc and metadata indexed in ODC DB, any thoughts?

eodatasets3/scripts/tostac.py
jeremyh: Oh bummer, I didn't realise it was used as a Python API already. I'll add backwards compat.

eodatasets3/scripts/tostac.py
jeremyh: I'm adding it below whenever at least one property uses `view`:

https://github.com/GeoscienceAustralia/eo-datasets/pull/98/files#diff-d9f5af2b32d41023013e97e5e6b86e646f6f4cd557c3b2ba32324bcfbfa98273R262-R264

My problem is that stac items fail validation if we declare `view` but don't have the required fields, and only some datasets add them.

jeremyh:
> Does changes in created field and instruments values will required updating existing STAC doc and metadata indexed in ODC DB, any thoughts?

I assume this shouldn't affect ODC-indexed datasets, as the original eo3 doc is unchanged. I don't know enough about our current use of stac outputs to know if the other change matters. Will it break anything that you know? 

That seems to be the common way they're represented in the stac samples I've seen, and how the instruments field is represented on Explorer's current stac output.

sachitrajbhandari:
> > Does changes in created field and instruments values will required updating existing STAC doc and metadata indexed in ODC DB, any thoughts?
> 
> I assume this shouldn't affect ODC-indexed datasets, as the original eo3 doc is unchanged. I don't know enough about our current use of stac outputs to know if the other change matters. Will it break anything that you know?
> 
> That seems to be the common way they're represented in the stac samples I've seen, and how the instruments field is represented on Explorer's current stac output.

Before indexing, we do STAC to DC metadata transform. So, we will need to update transform code for changes in `created` field and `instruments` values (https://github.com/opendatacube/odc-tools/blob/develop/libs/index/odc/index/stac.py)

alexgleith:
I think we should consider moving this tool out of EO Datasets and into the `odc-tools` repo here: https://github.com/opendatacube/odc-tools/blob/develop/libs/index/odc/index/stac.py

Then it can sit next to the STAC to ODC conversion script linked above.

jeremyh:
> I think we should consider moving this tool out of EO Datasets and into the `odc-tools` repo here: https://github.com/opendatacube/odc-tools/blob/develop/libs/index/odc/index/stac.py
> 
> Then it can sit next to the STAC to ODC conversion script linked above.


We might need a wider discussion around how we're handling libraries. eo-datasets is basically all metadata handling/validation/conversion, so if this should be moved maybe all of it should? But ODC-Tools is advertising itself as experimental code and everything we have that uses eo-datasets wants extreme stability (no accidental metadata or output changes for products<sup>1</sup>). The ODC-Tools repo doesn't run its tests or have docs setup, or versioning, and isn't installable from normal pypi like eodatasets, so I'm hesitant to rely on it from other repos at the moment... (should I be?)<sup>2</sup>

This PR is exposing functions for use by Explorer's stac 1.0.0 support (and fixing the validation failures), so it's going to depend on its stability quite heavily too. 

<sup>1</sup> eo-datasets came about years ago from frustration with our metadata code, as they were diffuse and inconsistent, and slight mistakes were accidentally introduced multiple times without people noticing for months on produced products. Its tests+versioning are strict.
<sup>2</sup> I'm completely biased of course, but short-term I'd rather move stac-to-eo3 here rather than the other way around if we want them side-by-side as we have all the infrastructure set up here (and other projects already depend on eodatasets through pypi). Or else move both of them up to odc-core, but that would slow down the release cycle (?). This is the only blocker on Explorer's stac 1.0 PR  so I'm excited by any solutions that are easy :slightly_smiling_face:

jeremyh:
Added a shim `dc_to_stac` function in the old place to maintain backwards compatibility.

I've also moved the API to `eodataset3.stac`, as the `eodatasets3.script` directory is more for CLI-frontends, not library APIs.

So it's now `eodatasets3.stac.to_stac_item()`

Added docstrings etc too, and have tried to rename the parameters to be more understandable.

jeremyh:
- Added a `verbose` option to `eo3-to-stac`, so people can tell what it's doing.
- Made schema downloads more robust
    - Use one session for all downloads
    - Use a timeout
    - Keep cached items much longer when they're stable Stac versions.
    - Allow skipping of the cache
    - Allow no-network-access usage

eodatasets3/scripts/tostac.py
jeremyh: Added shim fallback functions

whatnick (approved):
💯

whatnick (commented):
@alexgleith is better suited to review

alexgleith (approved):
Looks good to me.

As I said, it would be great to integrate this with the dataset assembler, so it's super simple for us to create STAC docs with out datasets!

PR #99, opened 2020-10-06
Add function to make a JPG from a singleband image

alexgleith:
The goal is to make it possible to product a simple thumbnail from a single image, rather than three images.

There's two ways of using this, either picking a single bit flag and making that white and everything else black, or providing a dict with a lookup table of bit flag to colour.


jeremyh:
Looks okay to me (once the build passes). Ideally we do these in-memory when possible, but the other thumbnail function doesn't currntly.

For most of the test data we're using tinier images (eg.  real datasets cropped to 20x20 pixels) to minimise bloating of the repo, and to make them fast when the extra data doesn't matter to the test.

alexgleith:
> Looks okay to me (once the build passes). Ideally we do these in-memory when possible, but the other thumbnail function doesn't currntly.

I tried fairly hard to use in-memory files, but I think they're failing in the other methods.

I'll look into the sample image and see if I can do something better. I'd like a more clear example of water to highlight, so that it's obvious that it works!

alexgleith:
Works now on a tiny image. Do you think we should actually compare the output to something, rather than just asserting that it's a file? Not sure what a good test is. We can't really test data, since the JPG ends up blurring things due to compression.

jeremyh:
> Works now on a tiny image. Do you think we should actually compare the output to something, rather than just asserting that it's a file? Not sure what a good test is. We can't really test data, since the JPG ends up blurring things due to compression.

That would be neat to do. A fuzzy-match pixel within a threshold (or something)? We're not doing it elsewhere in the repo, though we do try to check that it's loadable and of expected shape+properties. Eg https://github.com/GeoscienceAustralia/eo-datasets/blob/fab96dfc51e2755e46160fd5ce3a2e8a44d00a90/tests/integration/test_packagewagl.py#L460-L463 and https://github.com/GeoscienceAustralia/eo-datasets/blob/fab96dfc51e2755e46160fd5ce3a2e8a44d00a90/tests/integration/test_packagewagl.py#L469-L476

jeremyh:
The actual data is easier than thumbnails in this regard... in those cases we know the pixels shouldn't change at all, so we test their checksum:
https://github.com/GeoscienceAustralia/eo-datasets/blob/fab96dfc51e2755e46160fd5ce3a2e8a44d00a90/tests/integration/test_packagewagl.py#L433-L436

eodatasets3/images.py
```diff
@@ -686,6 +683,64 @@ def create_thumbnail(
...
+        lookup_table: Dict[int, Tuple[int, int, int]] = None,
+    ):
+        """
+        Write out a JPG thumbnail from a singleband image.
+        This takes in a path to a valid raster dataset and writes
+        out a file with only the values of the bit (integer) as white
+        """
+        assert not (
+            bit is not None and lookup_table is not None
+        ), "Please set either bit or lookup_table, and not both of them"
```
jeremyh: Assertions are quietly disabled if the user runs with `python -O` or `PYTHONOPTIMIZE=true`  (I've seen some Linux distros do this by default). It's uglier, but explicit is safer:
```suggestion
        if bit is not None and lookup_table is not None:
            raise ValueError(
                "Please set either bit or lookup_table, and not both of them"
            )
```

eodatasets3/images.py
```diff
@@ -686,6 +683,64 @@ def create_thumbnail(
...
+
+            with rasterio.open(interim_file, "w", **meta) as tmpdataset:
+                tmpdataset.write(out_data)
+            self.create_thumbnail(
+                [interim_file, interim_file, interim_file],
+                out_file,
+                static_stretch=stretch,
+            )
+        else:
+            temp_dir = Path("/tmp")
```
jeremyh: Use `tempfile` module like above to support all platforms.

eodatasets3/images.py
```diff
@@ -686,6 +683,64 @@ def create_thumbnail(
...
+
+                for value, rgb in lookup_table.items():
+                    for index in range(3):
+                        out_data[index][data == value] = rgb[index]
+
+        meta = dataset.meta
+        meta["driver"] = "GTiff"
+
+        if bit:
+            interim_file = Path(tempfile.gettempdir()) / "temp.tif"
```
jeremyh: I missed this before -- but this will use the same temp file if two processes run concurrently.

We use the `TemporaryDirectory` elsewhere to avoid this (it also cleans up automatically), eg:

https://github.com/GeoscienceAustralia/eo-datasets/blob/cbc21162134214b701c3ffb60e0e4ee8feb42cc7/eodatasets3/images.py#L552-L554

(Those parameters might be optional in this case. Using the output folder is to avoid errors of limited `/tmp` space when running huge parallel jobs on a machine.)

jeremyh (changes requested):
I haven't run it, but it looks good. Just a few one-liners.

eodatasets3/images.py
alexgleith: Oops, this is a holdover from testing. Thanks!

eodatasets3/images.py
alexgleith: Also good point.

alexgleith:
Can you turn on squash and merge, @jeremyh? It'll be good to get rid of the noise I created :-) 

(Or are you happy for me to?)

image

jeremyh:
I usually leave it off because people accidentally kill good history sometimes choosing the wrong one... but feel free to turn it on temporarily if it's easy.

(I usually squash locally and force-push to the branch, as it allows for more control. Such as only squashing certain commits.)

pre-commit also helps avoid formatting commits: https://github.com/GeoscienceAustralia/eo-datasets#pre-commit-setup

jeremyh (approved):

PR #100, opened 2020-10-14
Bump ard dataset 3.2.0

sixy6e:
Bumping the ARD dataset version number from 3.1.0 to 3.2.0

### DIFF
--- a/tests/integration/data/tostac/ga_ls8c_ard_3-1-0_088080_2020-05-25_final.odc-metadata.yaml
+++ b/tests/integration/data/tostac/ga_ls8c_ard_3-1-0_088080_2020-05-25_final.odc-metadata.yaml
@@ -3,7 +3,7 @@
 $schema: https://schemas.opendatacube.org/dataset
 id: 2aa69fcf-aa55-4747-9d95-3652f9fe79b0
 
-label: ga_ls8c_ard_3-1-0_088080_2020-05-25_final
+label: ga_ls8c_ard_3-2-0_088080_2020-05-25_final
 product:
   name: ga_ls8c_ard_3
   href: https://collections.dea.ga.gov.au/product/ga_ls8c_ard_3
@@ -71,7 +71,7 @@ properties:
   landsat:landsat_scene_id: LC80880802020146LGN00
   landsat:wrs_path: 88
   landsat:wrs_row: 80
-  odc:dataset_version: 3.1.0
+  odc:dataset_version: 3.2.0
   odc:file_format: GeoTIFF
   odc:processing_datetime: 2020-06-18 07:07:06.303867Z
   odc:producer: ga.gov.au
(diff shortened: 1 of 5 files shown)


tests/integration/test_tostac.py
```diff
@@ -11,10 +11,10 @@
 from tests.integration.common import run_prepare_cli
 
 TO_STAC_DATA: Path = Path(__file__).parent.joinpath("data/tostac")
-ODC_METADATA_FILE: str = "ga_ls8c_ard_3-1-0_088080_2020-05-25_final.odc-metadata.yaml"
+ODC_METADATA_FILE: str = "ga_ls8c_ard_3-2-0_088080_2020-05-25_final.odc-metadata.yaml"
 STAC_TEMPLATE_FILE: str = "ga_ls_ard_3_stac_item.json"
 STAC_EXPECTED_FILE: str = (
-    "ga_ls8c_ard_3-1-0_088080_2020-05-25_final.stac-item_expected.json"
+    "ga_ls8c_ard_3-2-0_088080_2020-05-25_final.stac-item_expected.json"
```
jeremyh, line 17: This refers to the test data files inside the repository, so the change to them is breaking Travis. We could rename the test data too, but I'd rather revert these lines as it's the correct version the test data was made from.

tests/integration/data/tostac/ga_ls8c_ard_3-1-0_088080_2020-05-25_final.odc-metadata.yaml
```diff
@@ -3,7 +3,7 @@
 $schema: https://schemas.opendatacube.org/dataset
 id: 2aa69fcf-aa55-4747-9d95-3652f9fe79b0
 
-label: ga_ls8c_ard_3-1-0_088080_2020-05-25_final
+label: ga_ls8c_ard_3-2-0_088080_2020-05-25_final
```
jeremyh, line 6: Same with these two files in the `tostac` folder. Can leave them as-is. 

They're previously-packaged files, not data running through the new packager.

jeremyh: And if we leave them as-is now, they wont show up on future find-replaces of similar version bumps :slightly_smiling_face:

jeremyh:
It still makes me slightly uneasy to hard-code the version like this, as anything could accidentally use the wrong eodatasets version against their input data.  

Could we maybe read the wagl version to know the dataset version, or something? (Or if preferred, throw error if wagl version isn't what's expected.)

tests/integration/test_tostac.py
sixy6e: no problem. I'll revert the commits.

PR #103, opened 2020-10-21
Don't Error if no h5py when generating thumbnails

omad:
Don't Error if no `h5py` when generating thumbnails


jeremyh (approved):
Thanks!

jeremyh:
This is a case we sadly don't have CI configured for... "do tests still pass without the optional dependencies installed"

PR #104, opened 2020-10-26
Packaging of wagl-water-atcor LMBSKYG products in the water-atcor-develop develop branch

Oceancolour-RG:
This commit focuses on the packaging of the wagl-water-atcor LMBSKYG product for Sentinel-2/MSI and Landsat-8/OLI products. Here:
1. The MNDWI image is used to mask land pixels, which are subsequently assigned as no-data;
2. expand_valid_data is set False to avoid the computation of complex geometry as a result of the MNDWI masking;
3. A simple work-around for Sentinel-2's absent level-1 metadata yaml was created;
4. A simple naming convention for the grid-names was created.


sixy6e:
Closing.  Appears polluted.

PR #105, opened 2020-10-26
Add in a lookuptable thumbnail function

alexgleith:
Add in a way to call the existing function for a singleband lookuptable from the dataset assembler class.

### DIFF
--- a/eodatasets3/assemble.py
+++ b/eodatasets3/assemble.py
@@ -180,6 +180,7 @@ class DatasetAssembler(EoFields):
         self._is_completed = False
         self._finished_init_ = True
 
+
     def _is_writing_files(self):
         """
         Have they written any files? Otherwise we're just writing a metadata doc
@@ -210,6 +211,12 @@ class DatasetAssembler(EoFields):
     @property
     def properties(self) -> StacPropertyView:
         return self._props
+
+    @property
+    def measurements(self) -> Dict[str: Tuple[GridSpec, Path]]:
+        return dict(
+            (name, (grid, path)) for grid, name, path in self._measurements.iter_paths()
+        )
 
     @property
     def label(self) -> Optional[str]:
@@ -782,6 +789,14 @@ class DatasetAssembler(EoFields):
     def _crs_str(self, crs: CRS) -> str:
         return f"epsg:{crs.to_epsg()}" if crs.is_epsg_code else crs.to_wkt()
 
+    def _document_thumbnail(self, thumb_path, kind=None):
+        self._checksum.add_file(thumb_path)
+
+        accessory_name = "thumbnail"
+        if kind:
+            accessory_name += f":{kind}"
+        self.add_accessory_file(accessory_name, thumb_path)
+
     def write_thumbnail(
         self,
         red: str,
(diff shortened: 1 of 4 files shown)


omad (dismissed):
Looks like an improvement to me, pushing more code in here from `datacube-alchemist`.

My only concern, and not a blocker, is that this doesn't go far enough. It's still too method call driven which means extra logic in the client. Maybe we should add parameters to `DatasetAssembler.write_thumbnail()` and it gets to decide on the method to use to generate the thumbnail...

alexgleith:
What do you think, @jeremyh? Want me to weave the two functions into a single meta function with two code paths?

omad (dismissed):

eodatasets3/assemble.py
```diff
@@ -846,12 +857,52 @@ def write_thumbnail(
...
-        accessory_name = "thumbnail"
-        if kind:
-            accessory_name += f":{kind}"
-        self.add_accessory_file(accessory_name, thumb_path)
+        self._document_thumbnail(thumb_path, kind)
+
+    def write_thumbnail_singleband(
+        self,
+        measurement: str,
+        thumb_path,
```
jeremyh: This argument may have been for debugging? (it's never used, being overridden in the first line. And the paired `create_thumbnail` function doesn't have it either)

eodatasets3/assemble.py
```diff
@@ -846,12 +857,52 @@ def write_thumbnail(
...
+        Use a bit (int) as the value to scale from black to white to
+        i.e., 0 will be BLACK and bit will be WHITE, with a linear scale between.
+        OR
+        Provide a lookuptable (dict) of int (key) [R, G, B] (value) fields
+        to make the image with.
+        """
+
+        thumb_path = self.names.thumbnail_name(self._work_path, kind=kind)
+
+        _, image_path = self.measurements.get(measurement, None)
```
jeremyh: The unpacking will always fail when measurement is unknown, so we'll never reach the below validation.

```suggestion
        _, image_path = self.measurements.get(measurement, (None, None))
```

jeremyh (commented):
Looks good. It's different enough of a situation that I don't mind a second function for convenience. Though I like where Damien's coming from.

If combining them means making all the current red/green/blue args optional on the existing function I'd probably prefer not to combine, as their arguments are too different and we'll have a confusing set of completely-mutually-exclusive optional/required parameter combinations on the one function.

--- 

Perhaps also add a line to one of our minimal tests to run the function, eg:

```
p.write_thumbnail_singleband('blue', None, bit=1)
```
to https://github.com/GeoscienceAustralia/eo-datasets/blob/224f97410995e083982fa20ce75fe8e069f8fdbf/tests/integration/test_assemble.py#L200

(I also ran it at first without any `bit` argument and got a cryptic error message -- we're never checking the `bit is None and lookup_table is None` case in our old method parameter checking)

jeremyh:
Bike shedding... should it be `write_thumbnail_singleband` or `write_thumbnail_single_band` to have consistent naming style? (we have, for example `self.write_measurements_odc_xarray()` elsewhere)

alexgleith:
> Bike shedding... should it be `write_thumbnail_singleband` or `write_thumbnail_single_band` to have consistent naming style? (we have, for example `self.write_measurements_odc_xarray()` elsewhere)

I think singleband is a single word... keeping it as it is ;-)

omad:
Install `pre-commit` Alex!

omad (approved):
👯

jeremyh (approved):
👍

PR #107, opened 2020-10-26
Add Option to Inherit the Valid Geometry from Source Datasets

omad:
While running some processing on derivative data, we discovered that in some cases 
it was taking upwards of an hour to perform the vector operations required to derive 
the valid data geometry.

In many cases it should be possible to copy the geometry from the input dataset and short
circuit the need to do any calculations at all.

This PR adds a flag to `.add_dataset_source()` called `inherit_geometry` which makes this
happen.

I've also added some files to ease testing, and perhaps move to new setup in CI. If Docker
is installed you can now run `make docker-test` which will build a new docker image and
run the tests against an installed version of **eo-datasets**.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,52 @@
+FROM opendatacube/geobase:wheels as env_builder
+ARG py_env_path=/env
+ARG ENVIRONMENT=test
+
+COPY requirements*.txt /tmp/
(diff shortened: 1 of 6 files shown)


jeremyh (approved):
Looks good. I like the included test.

alexgleith (approved):
Looks good! Nice and simple :-)

PR #117, opened 2020-12-22
Use rel=collection in stac, not rel=parent

jeremyh:
As is strongly encouraged by the spec: https://github.com/radiantearth/stac-spec/blob/4988a40f6071666d345e4a1c5adc9a6338973769/item-spec/item-spec.md#relation-types

Users can specify either a collection_url or an explorer_url to have the collection url added to their Item.

---
Includes an unrelated clean-up I had done previously (yes, this is unwise): move the stac property conversion logic into its own method.


alexgleith (approved):
Very nice!

Issue #110, opened 2020-10-27
Move away from Travis CI


jeremyh:
Travis queues are getting longer and failing more often. Our other projects have moved to Github Actions so this project should too.

We accidentally merged a PR with a (trivial) test failure yesterday, because it was merged manually to work around a Travis failure.

jeremyh:
Done in #120 and #121

PR #120, opened 2021-02-24
Move to Github Actions

jeremyh:
The Travis build is broken again, and rather than fix it, we're overdue for a move. Every other repo has moved to Github CI already.

For #110

---

Also, removing the ancient S2 prepare scripts which were carried over from eodatasets1. They had their own test failure under docker which I suspect were from different library versions.


omad:
Hooray! Nice work Jeremy, and thank you!

Issue #123, opened 2021-02-25
Overview image in generated COGs use default 128 pixel tile size


Kirill888:
Default tile size for overview images is still stuck in the 90s, 128x128 pixel.

```
tiffinfo ga_ls_wo_3_101072_2019-12-13_final_water.tif | grep "Tile Width"
  Tile Width: 512 Tile Length: 512
  Tile Width: 128 Tile Length: 128
  Tile Width: 128 Tile Length: 128
  Tile Width: 128 Tile Length: 128
```

This is way too small given that our overview images can be as big as 4800x4800 pixels.

Overview tile size is configured via GDAL_TIFF_OVR_BLOCKSIZE= parameter, see here for example

https://github.com/opendatacube/datacube-core/blob/afca22831eecf8092e0c293064a9ea465aa6d778/datacube/utils/cog.py#L144

I think this code should be using write_cog from datacube lib.
https://github.com/GeoscienceAustralia/eo-datasets/blob/05c5e812ecf2e8baaa8a3e33cc55717dd452b46d/eodatasets3/images.py#L440-L448

https://datacube-core.readthedocs.io/en/latest/dev/api/generate/datacube.utils.cog.write_cog.html

jeremyh:
Thanks Kirill. Datacube's COG package didn't exist when we added these methods, but they clearly look more thorough. I agree we should move to it.

Kirill888:
They do work on xarrays though, not raw numpy arrays. Also not sure how attached you are to special handling HDF data sources. Other thing done differently in core is that first pass compression is disabled, which is faster, but uses more RAM, so something to be aware of. It's been on my todo list to enable first pass compression when needed (large mosaics is one such example).

jeremyh:
Closing, as the default was changed.

PR #124, opened 2021-02-26
Increase default overview block size to 512

jeremyh:
The default is only 128. Includes test to verify.
.
As reported in #123


alexgleith (approved):
Extremely nice!

PR #129, opened 2021-03-11
Remove unnecessary read of images in note measurements 

kieranricardo:
This PR changes `DatasetAssembler.note_measurement` to not read the image file if the `expand_valid_data` parameter is `False`. This optimization reduces file IO which can be a bottleneck especially on NCI.


jeremyh (approved):
Thanks @kieranricardo

PR #130, opened 2021-03-14
Update S2 L1 YAML configuration to support EO3

ohehir:
Current S2 L1C products (at fj7) don’t have an associated EO3 yaml file (for S2 collection update processing/packaging); there are associated old EO style yaml file for current S2 collection processing. 

This PR updates eodatasets3 to include S2 L1 YAML configuration to support EO3.


tests/integration/prepare/test_prepare_sinergise_sentinel_l1.py
```diff
@@ -0,0 +1,194 @@
...
+import yaml
+import datetime
+
+import sys
+
+sys.path.append("/g/data/up71/projects/index-testing-wagl/repo/eo-datasets/tests")
+sys.path.append("/g/data/up71/projects/index-testing-wagl/repo/eo-datasets/eodatasets3")
+sys.path.append(
+    "/g/data/up71/projects/index-testing-wagl/repo/eo-datasets/eodatasets3/prepare"
+)
```
jeremyh:
```suggestion
```

Remove these leftover lines from dev

tests/integration/prepare/test_prepare_sinergise_sentinel_l1.py
```diff
@@ -0,0 +1,194 @@
+# from eodatasets3.prepare import sentinel_sinergise_L1_prepare
```
jeremyh: Remove commented-out code to keep it clean

```suggestion
```

tests/integration/prepare/test_prepare_sinergise_sentinel_l1.py
```diff
@@ -0,0 +1,194 @@
...
+sys.path.append(
+    "/g/data/up71/projects/index-testing-wagl/repo/eo-datasets/eodatasets3/prepare"
+)
+
+from common import check_prepare_outputs  # Change to tests.common
+from common import run_prepare_cli  # Change to tests.common
+
+from prepare import sentinel_sinergise_L1_prepare  # eodatasets3.prepare
+
+# from sentinel_sinergise_L1_prepare import run
```
jeremyh:
```suggestion
```

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,225 @@
...
+    "09": "water_vapour",
+    "10": "swir_1_cirrus",
+    "11": "swir_2",
+    "12": "swir_3",
+}
+
+
+def extract_metadata_from_product_info(product_path: Path) -> Dict:
+    fp = open(product_path)
+    product = json.loads(fp.read())
```
jeremyh: We should close the `fp`.

(a `with` statement is ideal, similar to how the yaml is read elsewhere in this PR)

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,225 @@
...
+            if ".jp2" in dataset and "preview" not in dataset and "TCI" not in dataset:
+                band = dataset.replace(".jp2", "").replace("B", "")
+                name = SENTINEL_MSI_BAND_ALIASES[band]
+                path = ds_path / dataset
+                p.note_measurement(path=path, name=name)
+
+        return p.done()
+
+
+@click.command(help=__doc__)
```
jeremyh, line 164: The `__doc__` is the comment at the top of the file, but we don't have one here. See an example in the [Landsat packager](https://github.com/GeoscienceAustralia/eo-datasets/blob/ad388f10bfc967ac2204f7fc1f76d536e35b4b2c/eodatasets3/prepare/landsat_l1_prepare.py#L1-L5).

These `help=` properties in click are what is shown when running `--help` on the command-line.

It's all blank at the moment:
```
❯ eo3-prepare sin-sentinel-l1 --help
Usage: eo3-prepare sin-sentinel-l1 [OPTIONS] PRODUCT_PATH METADATA_XML_PATH
                                   FORMAT_CORRECTNESS_PATH OUTPUT_YAML_PATH
                                   DS_PATH

Options:
  --help  Show this message and exit.

```

See the Landast one for an example (it has `help=` on [its options](https://github.com/GeoscienceAustralia/eo-datasets/blob/c115ee1542f3a4a43f9d076aaec4a3f021692d8b/eodatasets3/prepare/landsat_l1_prepare.py#L351)):

```
❯ eo3-prepare landsat-l1 --help
Usage: eo3-prepare landsat-l1 [OPTIONS] [DATASETS]...

  Prepare eo3 metadata for USGS Landsat Level 1 data.

  Input dataset paths can be directories or tar files.

Options:
  --output-base DIRECTORY         Write metadata files into a directory
                                  instead of alongside each dataset

  --source PATH                   Path to the source telemetry data for all
                                  of the provided datasets(either the folder
                                  or metadata file)

  --producer TEXT                 Organisation that produced the data:
                                  probably either 'ga.gov.au' or 'usgs.gov'.

  --overwrite-existing / --skip-existing
                                  Overwrite if exists (otherwise skip)
  --newer-than DATE               Only process files newer than this date
  --help                          Show this message and exit.

```

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,225 @@
...
+    type=str,
+)
+@click.argument(
+    "format_correctness_path",
+    type=str,
+)
+@click.argument(
+    "output_yaml_path",
+    type=str,
+)
```
jeremyh: It can be quite error prone to take so many unnamed arguments like this with an exact required order. It wont be  obvious from looking at the command whether they are in the correct order needed.
              
The other packagers [are using](https://github.com/GeoscienceAustralia/eo-datasets/blob/eodatasets3/eodatasets3/prepare/landsat_l1_prepare.py#L335-L347)  `click.option` instead of `click.argument` for similar things, so they are named explicitly:

Eg. 
`--product-info /one/path --output /other/path --format-correctness /third/path` 
instead of just `/one/path /other/path /third/path`

---

But also, do all of these files have predictable names (like `productInfo.xml`)? If so, the packager could just find them itself in a single folder. It's one less thing that the user will need to do themselves. This is lower priority than making them options.

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,224 @@
...
+def extract_metadata_from_product_info(product_path: Path) -> Dict:
+    fp = open(product_path)
+    product = json.loads(fp.read())
+
+    synergise_product_name = product["name"]
+    synergise_product_id = product["id"]
+    timestamp = product["tiles"][0]["timestamp"]
+    utm_zone = product["tiles"][0]["utmZone"]
+    latitude_band = product["tiles"][0]["latitudeBand"]
+    grid_square = product["tiles"][0]["gridSquare"]
```
jeremyh: Are we sure that there's always only one tile in the list?

Perhaps add a safety check so we don't risk quietly missing data in the future:
```suggestion
    if len(product["tiles"]) > 1:
        raise NotImplementedError("No support for multi-tiled products yet")
    [tile] = product["tiles"]
    timestamp = tile["timestamp"]
    utm_zone = tile["utmZone"]
    latitude_band = tile["latitudeBand"]
    grid_square = tile["gridSquare"]
```

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,224 @@
...
+    region_code = "%s%s%s" % (utm_zone, latitude_band, grid_square)
+
+    return {
+        "synergise_product_name": synergise_product_name,
+        "synergise_product_id": synergise_product_id,
+        "timestamp": timestamp,
+        "region_code": region_code,
+    }
```
jeremyh: It's not high priority, but I'd consider having these (and other similar) dicts use the the real property names ("eo:platform" etc), so that we don't have yet another set of names for every property to convert back and forth between.

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,224 @@
...
+):
+    product_info = extract_metadata_from_product_info(Path(product_path))
+    metadata_xml = extract_metadata_from_metadata_xml(metadata_xml_path)
+    format_correctness = extract_metadata_from_format_correctness(
+        format_correctness_path
+    )
+
+    uuid, path = prepare_and_write(
+        Path(output_yaml_path),
+        Path(ds_path),
```
jeremyh: Use `type=PathPath()` in Click so we don't have to convert them manually.

[Example](https://github.com/GeoscienceAustralia/eo-datasets/blob/c115ee1542f3a4a43f9d076aaec4a3f021692d8b/eodatasets3/prepare/landsat_l1_prepare.py#L339)

(`from eodatasets3.ui import PathPath`)

tests/integration/prepare/test_prepare_sinergise_sentinel_l1.py
```diff
@@ -0,0 +1,177 @@
...
+            "datetime": datetime.datetime(
+                2020, 10, 11, 0, 6, 49, 882000, tzinfo=datetime.timezone.utc
+            ),
+            "eo:cloud_cover": 24.9912,
+            "eo:gsd": 10,
+            "eo:instrument": "MSI",
+            "eo:platform": "sentinel-2a",
+            "eo:sun_azimuth": 37.3713908882192,
+            "eo:sun_elevation": 46.3307328858312,
+            "odc:dataset_version": "1.0.0",
```
jeremyh: I'd consider putting the processing time in the version, [like we do with Landsat](https://github.com/GeoscienceAustralia/eo-datasets/blob/c115ee1542f3a4a43f9d076aaec4a3f021692d8b/eodatasets3/prepare/landsat_l1_prepare.py#L313)... unless Sinergise have a specific way of tracking new dataset versions?

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,224 @@
...
+        p.properties["sentinel:datastrip_id"] = metadata_xml["datastrip_id"]
+
+        directory = [f for f in listdir(ds_path) if isfile(join(ds_path, f))]
+
+        for dataset in directory:
+            if ".jp2" in dataset and "preview" not in dataset and "TCI" not in dataset:
+                band = dataset.replace(".jp2", "").replace("B", "")
+                name = SENTINEL_MSI_BAND_ALIASES[band]
+                path = ds_path / dataset
+                p.note_measurement(path=path, name=name)
```
jeremyh, line 159: It would be nice to use `p.note_accessory()` too, to add the the original `xml` and other files. 

Ideally every useful file in their package is included: either as a measurement or as an accessory file.

jeremyh (changes requested):
Nice work Aaron, nice and clean. I like the test :)

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
ohehir: I've made them all options.

That is a good idea re finding the inputs automatically.

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
ohehir: done

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
ohehir: done

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
ohehir: added

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
ohehir: help had been added

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,219 @@
...
+from os.path import isfile, join
+import uuid
+import click
+from typing import Dict, Tuple
+from eodatasets3 import DatasetAssembler
+from eodatasets3.ui import PathPath
+
+"""
+Prepare eo3 metadata for Sentinel-2 Level 1C data produced by Sinergise.
+"""
```
jeremyh: A docstring has to be the first thing in the file (above the imports) or it wont work.

You can test if it's working by running `--help` on your command (notice this text isn't included):

```
❯ eo3-prepare sin-sentinel-l1 --help
Usage: eo3-prepare sin-sentinel-l1 [OPTIONS]

Options:
  --product PATH             Path to productInfo.json in sinergise dataset
                             [required]

  --metadata_xml TEXT        Path to metadata.xml in sinergise dataset
                             [required]

  --format_correctness TEXT  Path to FORMAT_CORRECTNESS.xml in sinergise
                             dataset  [required]

  --dataset_document PATH    Path to output dataset document (yaml)
                             [required]

  --dataset PATH             Path to sinergise dataset  [required]
  --help                     Show this message and exit.
```

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,219 @@
...
+    help="Path to metadata.xml in sinergise dataset",
+)
+@click.option(
+    "--format_correctness",
+    type=str,
+    required=True,
+    help="Path to FORMAT_CORRECTNESS.xml in sinergise dataset",
+)
+@click.option(
+    "--dataset_document",
```
jeremyh: We should use dashes as separators in the names to stay consistent with other commands

(Click will still give them to your function with underscores)

```suggestion
    "--metadata-xml",
    type=str,
    required=True,
    help="Path to metadata.xml in sinergise dataset",
)
@click.option(
    "--format-correctness",
    type=str,
    required=True,
    help="Path to FORMAT_CORRECTNESS.xml in sinergise dataset",
)
@click.option(
    "--dataset-document",
```

jeremyh (commented):
:+1:

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
```diff
@@ -0,0 +1,219 @@
...
+)
+@click.option(
+    "--metadata_xml",
+    type=str,
+    required=True,
+    help="Path to metadata.xml in sinergise dataset",
+)
+@click.option(
+    "--format_correctness",
+    type=str,
```
jeremyh: I noticed these two are `type=str`, but the other paths are `type=PathPath`. Is there a reason? Maybe make them all Paths?

ohehir: Nope, they should be string

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
ohehir: I've moved the doc string to the top

eodatasets3/prepare/sentinel_sinergise_L1_prepare.py
jeremyh: Why?

jeremyh: Is it because `minidom.parse()` didn't work with a Path object? 

You can convert to string when doing the minidom call: `minidom.parse(str(path))`

(... otherwise it can be a pain to have to work with a mixture of inconsistent types all throughout the code.)

jeremyh (approved):
Looks good. Nice work Aaron.

(I'm not a big fan of inconsistencies in having both only-string and only-pathlib variables mixed throughout one file [as mentioned above]. But those variables are localised to this file and so it's only a minor quibble. Nice work on adding types everywhere though, so it _is_ actually clear what they are, despite the similar `*_path` naming)

PR #131, opened 2021-03-19
Expand wagl packager for S2

jeremyh:
Includes an integration test with the sample dataset provided to us.

This also includes improvements to the core Assembler that are useful for Sentinel:
- Improve the grid name generation: it can now name grids by resolution, or if nothing else works, alphabetically (grid 'a', grid 'b' etc). This was desired because Sentinel has a more diverse set of grids.
- Allow L1 provenance to be optional.
- Fix spurious "property is being overridden" warnings, because it wasn't normalising both values before comparing.
- The "unknown property" warning now gives a [filled-out Github issue template](https://github.com/GeoscienceAustralia/eo-datasets/pull/131/commits/5059f6d56c657844a65c9a4168b74f0e4dc77eb0#diff-beca4612399753d331db3f56c4560edbda7b54d67a717aca8ba775a5ba386b60R525-R543) to raise an issue about it easily

Remaining:
- [ ] Confirm and switch on the S2-specific naming conventions
- [ ] Add cloud-cover? (in Landsat, it comes from L1)
- [ ] Update maturity handling logic to mark NRT when needed
- [ ] Get a decision on what OA resolution to use. Currently S2 defaults to 20m, but @robbibt has proposed 10m.
- [ ] Confirm and update the Level 1 metadata so they gel together (clear the warnings L1 is generating)
- [ ] Add Level 1 provenance to the tests (now that #130 is merged.)
- [ ] Is there other Sentinel 2 stac metadata we should be including? Look at existing Stac products


jeremyh:
This has useful enough, stand-alone additions that I think the remaining tasks can be separate PRs.

PR #134, opened 2021-03-26
ARD maturity should always be NRT when Level 1 is NRT

jeremyh:


jeremyh:
CI failure is unrelated to this PR -- it looks like the test added for Synergise is unstable (it depends on the order of an unordered set of elements).

PR #135, opened 2021-03-29
Sentinel 2 - Level 1C eo3 prepare scripts

ohehir:
Merged sinergise and esa level 1C eo3 prepare scripts into a single script.


eodatasets3/prepare/sentinel_L1C_prepare.py
```diff
@@ -0,0 +1,393 @@
+"""
```
jeremyh: Keep py file names lowercase, for consistency. 

(I'm surprised the pep8 linter doesn't complain about this)

eodatasets3/prepare/sentinel_L1C_prepare.py
```diff
@@ -0,0 +1,393 @@
...
+            "datastrip_metadata": datastrip_metadata,
+        }
+
+
+def prepare_and_write(
+    dataset: Path,
+    dataset_document: Path,
+) -> Tuple[uuid.UUID, Path]:
+    # Process esa dataset
+    if "zip" in str(dataset):
```
jeremyh: I'm a bit iffy about using 'zip' as the way to distinguish the type of dataset (ESA vs Synergise). From experience, people often try to run these things with unzipped folders or otherwise, and it's going to quietly try to parse all the metadata differently depending on whether someone has done that.

Is there a file inside that we could use to detect? Or a more specific part of the filename?

eodatasets3/prepare/sentinel_L1C_prepare.py
```diff
@@ -0,0 +1,393 @@
...
+
+        return {
+            "synergise_product_name": synergise_product_name,
+            "synergise_product_id": synergise_product_id,
+            "timestamp": timestamp,
+            "region_code": region_code,
+        }
+
+
+def process_metadata_xml(metadata_xml_path):
```
jeremyh: All the type hints from the old prepare script seem to be removed -- across all the functions in this file.

Eg, the old one: https://github.com/GeoscienceAustralia/eo-datasets/blob/eodatasets3/eodatasets3/prepare/sentinel_sinergise_L1_prepare.py#L56

(they're especially useful here because this file has a mix of `str` and `Path` variables that are all called "path", and this can be confusing to people. .... ideally, [make them all Paths](https://github.com/GeoscienceAustralia/eo-datasets/pull/130#discussion_r595605063)!)

eodatasets3/prepare/sentinel_L1C_prepare.py
```diff
@@ -0,0 +1,393 @@
...
+
+            return p.done()
+
+
+@click.command(help=__doc__)
+@click.option(
+    "--dataset",
+    type=PathPath(),
+    required=True,
+    help="Path to ESA zipped dataset or Sinergise dataset directory",
```
jeremyh: Nice, specific docs :)

eodatasets3/scripts/prepare.py
```diff
@@ -13,7 +13,7 @@ def run():
 
 
 run.add_command(landsat_l1, name="landsat-l1")
-run.add_command(sin_sentinel_l1, name="sin-sentinel-l1")
+run.add_command(sentinel_l1c, name="sentinel-l1C")
```
jeremyh, line 16:
```suggestion
run.add_command(sentinel_l1c, name="sentinel-l1")
```

Keep lowercase, like the others. And we probably don't need the `C`. 

... Landsat is just `landsat-l1`, so we my as well match it.

jeremyh (changes requested):
Thanks Aaron.

- We probably need to review the sentinel stac properties, but I might do this in a separate PR. (... some of these look more like ODC properties, in the `sentinel:` namespace [?] And others like 'orbit' probably have general standards already? We should avoid making up our own in anothers' namespace)
- There's a stray swap file included in a commit: https://github.com/GeoscienceAustralia/eo-datasets/blob/eabdc479958443ad7ed17f339ba9ccb75af62f09/eodatasets3/prepare/.landsat_l1_prepare.py.swp

(you can add `*,swp` to you `.gitignore` file to avoid accidentally committing these in the future, if your editor uses such names)

jeremyh:
- Did a review of our S2 properties here instead of as a new PR. Made them a little more consistent with our other products, and external Sentinel 2 stac products. We don't want to invent our own properties if we can help it.
- Updated the inheritable properties so that wagl product will get them by default too.

jeremyh:
There'll probably be more review before we finalise our properties, but this is good enough for develop.

jeremyh (approved):

PR #136, opened 2021-03-31
Digital Earth Africa scene_id changes

eloise-b:
Adding to assemble.py and properties.py to include the line:
"landsat:scene_id": None,
This allows for the format of DEAfrica's metadata so that the scene id can be found.
And to model.py the change is adding a line so that the scene_id and information for DEAfrica are looked for.
I think these answer what are commented as TODO for allowing DEAfrica?


eloise-b:
@jeremyh @alexgleith

alexgleith:
Can you please add a description of the changes?

Hey @jeremyh does this have any further ramifications?

eloise-b:
> Can you please add a description of the changes?

Adding to assemble.py and properties.py to include the line:
`"landsat:scene_id": None,`
This allows for the format of DEAfrica's metadata so that the scene id can be found.
And to model.py the change is adding a line so that the scene_id and information for DEAfrica are looked for. 
I think these answer what are commented as TODO for allowing DEAfrica?

jeremyh:
This is another case of us slightly deviating from newer Stac standard: we currently use the property name `landsat:landsat_scene_id` in our collection 3 products, as that's what Stac examples used at the time we made them (it matches exactly the field name in USGS's MTL files), but newer Stac examples use `landsat:scene_id`, which is what Eloise is adding here.

_Usually_ we try to make eodatasets normalise these sorts of things so all products are consistent -- such as automatically converting `landsat:scene_id` to our existing field choice (as well as adding the reverse in the Stac converter).

... but if products are already published in Africa with this field name we may need to include the inconsistency

alexgleith:
Can you look into the linting and test failures please @eloise-b

jeremyh:
Done in #166

PR #138, opened 2021-04-09
Add support for USGS C2 to Landsat L1 prepare

dunkgray:
Adding support for L1 C2.
Note, I don't really know the implications of this change;
  eodatasets3/properties.py 
"odc:dataset_maturity": None,
So a different approach might be needed.


jeremyh:
Ooh, the `odc:dataset_maturity` warning is correctly finding a bug (I guess we haven't run RT data before?): the actual property name should be `dea:dataset_maturity`

eodatasets3/prepare/landsat_l1_prepare.py
```diff
@@ -235,23 +284,26 @@ def prepare_and_write(
...
+    mtl_doc, root_element, mtl_filename = get_mtl_content(ds_path)
     if not mtl_doc:
         raise ValueError(f"No MTL file found for {ds_path}")
-
-    usgs_collection_number = mtl_doc["metadata_file_info"].get("collection_number")
+    coll = "C2" if root_element == "landsat_metadata_file" else "C1"
+    coll_map = LANDSAT_MTL_MAP[coll]
+    usgs_collection_number = mtl_doc[coll_map["product_contents_cn"]].get(
+        "collection_number"
+    )
```
jeremyh: I'm slightly iffy about this first `if` statement, as more collections will be released in the future (... and already, pre-collection) that will quietly identify as the wrong collection number from it.

The `usgs_collection_number` below it contains the same information (but as an int), as is calculated from actual properties, so maybe we could just reuse that instead of this extra `coll` variable? (it also has a clearer name :slightly_smiling_face:)

eodatasets3/properties.py
```diff
@@ -234,6 +234,7 @@ class StacPropertyView(collections.abc.Mapping):
         "odc:producer": producer_check,
         "odc:product_family": None,
         "odc:region_code": None,
+        "odc:dataset_maturity": None,
```
jeremyh: As mentioned in the previous comment -- this was an existing bug (not introduced by you), and we should use `dea:dataset_maturity` instead of adding this one.

eodatasets3/prepare/landsat_l1_prepare.py
```diff
@@ -452,6 +519,9 @@ def _normalise_dataset_path(input_path: Path) -> Path:
         input_path = input_path.parent
 
     mtl_files = list(input_path.rglob("*_MTL*"))
+
+    if len(mtl_files) > 1:
+        mtl_files = list(input_path.rglob("*_MTL.txt"))
```
jeremyh: It took me a moment to understand this addition. I think we should just add `.txt` to the first statement. It only expects a text file from it, and will fall over otherwise

jeremyh (commented):
Wow, this follows existing conventions well, is clean, doesn't duplicate code, and comes with a test. Nice work @dunkgray!

A few _very_ minor comments.

eodatasets3/prepare/landsat_l1_prepare.py
dunkgray: Excellent, done.

eodatasets3/properties.py
dunkgray: Nice spotting.  Fixed.

eodatasets3/prepare/landsat_l1_prepare.py
dunkgray: Yes, if there are more collections the if statement will need to change.   Using usgs_collection_number is tough since it relies on the collection_key to get its value.  I've changed coll to collection_key, to use a clearer name.

dunkgray:
Thanks Jeremy! I try not to drag the standard of your code down :)

jeremyh (approved):
:+1:

PR #143, opened 2021-04-20
Dockerfile clean-up & dependency pin

jeremyh:
- Reorder docker file for better caching (it shouldn't need to re-install dependencies every time
code changes)
- Improved dependency handling (using pip-tools)
  - Generate dependencies from setup.py instead of duplicating them (... and forgetting some)
  - Pin dependencies for build stability.
- Avoid leaving dev files and dependencies in non-dev builds
- Clean-up some Dockerfile lint warnings and "best practices"
- Cache pre-commit in dev docker images, so that the environment isn't rebuilt every time the linter is run.
- Improve, update the repository README.md, and mention Docker.

Dependencies can be updated with `make dependency-update`, which will refreeze them within the docker container's python installation.


jeremyh:
... a strange pytest error that I'm not seeing in my local Docker container.

edit: fixed. Running pytest-coverage on a mounted volume as a non-root user seemed to have strange issues on Github CI (no "permission denied", yet some read/writes fail?).

dunkgray (approved):

PR #145, opened 2021-04-21
C2wagl

dunkgray:
Adding some tests.  Get wagl to find the metadata yaml within the directory.
Update the collection and dataset version number.

### DIFF
--- a/eodatasets3/wagl.py
+++ b/eodatasets3/wagl.py
@@ -1,7 +1,7 @@
 """
 Package WAGL HDF5 Outputs
 
-This will convert the HDF5 file (and sibling fmask/gqa files) into
+This converts the HDF5 file (and sibling fmask/gqa files) into
 GeoTIFFS (COGs) with datacube metadata using the DEA naming conventions
 for files.
 """
--- a/eodatasets3/wagl.py
+++ b/eodatasets3/wagl.py
@@ -590,7 +590,7 @@ def package(
                 p.platform, p.producer, p.properties.get("landsat:collection_number")
             )
 
-            p.dataset_version = f"{org_collection_number}.2.0"
+            p.dataset_version = f"{org_collection_number}.2.1"
             p.region_code = _extract_reference_code(p, granule.name)
 
             _read_gqa_doc(p, granule.gqa_doc)
--- a/eodatasets3/wagl.py
+++ b/eodatasets3/wagl.py
@@ -359,6 +359,8 @@ def _get_level1_metadata_path(wagl_doc):
         if not level1_metadata_path.exists():
             # Assume the yaml is in the source level1 dir
             [level1_metadata_path] = source_level1.rglob("*.odc-metadata.yaml")
+    else:
+        level1_metadata_path = None
     return level1_metadata_path
 
 
--- a/eodatasets3/wagl.py
+++ b/eodatasets3/wagl.py
@@ -350,6 +350,18 @@ def do(name: str, heading=False, **fields):
         secho("(done)")
 
 
+def _get_level1_metadata_path(wagl_doc):
+    source_level1 = Path(get_path(wagl_doc, ("source_datasets", "source_level1")))
+    # The odc-meta data can be in the source-level or alongside it.
+    if source_level1.exists():
+        # if source_level1.isdir():
+        level1_metadata_path = source_level1.with_suffix(".odc-metadata.yaml")
+        if not level1_metadata_path.exists():
+            # Assume the yaml is in the source level1 dir
+            [level1_metadata_path] = source_level1.rglob("*.odc-metadata.yaml")
+    return level1_metadata_path
+
+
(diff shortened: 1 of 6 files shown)


eodatasets3/wagl.py
```diff
@@ -350,6 +350,19 @@ def val(v: Any):
...
 
 
+def _get_level1_metadata_path(wagl_doc):
+    source_level1 = Path(get_path(wagl_doc, ("source_datasets", "source_level1")))
+    # The odc-metadata can be in the source directory or alongside it.
+    if source_level1.exists():
+        level1_metadata_path = source_level1.with_suffix(".odc-metadata.yaml")
+        if not level1_metadata_path.exists():
+            # Assume the yaml is in the source level1 dir
+            [level1_metadata_path] = source_level1.rglob("*.odc-metadata.yaml")
```
jeremyh: I'm iffy about this. There can be many level 1 scenes in one directory (as they're usually zip files or tar files), and when a metadata file is missing or deleted, this will quietly grab another arbitrary one in the same directory. Which will be very wrong.

Currently we have a "standard" naming convention of metadata being the identical filename to the input file, with an added `.odc-metadata.yaml` suffix.

 .... and if users want to put their metadata elsewhere, or use non-standard naming, they then just specify the level 1 path explicitly to the packager: `--level1 /some/path/to/yaml`.

Is there are third option we need to solve here? It sounds like you have files with completely different names in sub-directories?

jeremyh: We _really_ don't want to quietly link to the wrong source provenance, so I'm hesitant to make this level1 scanning too loose and magic.

dunkgray: Thanks Jeremy.  I've dialled back the magic.

eodatasets3/wagl.py
```diff
@@ -350,6 +350,20 @@ def val(v: Any):
...
 
+def _get_level1_metadata_path(wagl_doc):
+    source_level1 = Path(get_path(wagl_doc, ("source_datasets", "source_level1")))
+    # The odc-metadata can be in the source directory or alongside it.
+    if source_level1.exists():
+        level1_metadata_path = source_level1.with_suffix(".odc-metadata.yaml")
+        if not level1_metadata_path.exists():
+            # Assume the yaml is in the source level1 dir
+            metadata_yaml = source_level1.name + ".odc-metadata.yaml"
+            level1_metadata_path = source_level1 / metadata_yaml
```
jeremyh, line 361: This has slightly strange fallback behaviour if it doesn't exist. We really only want that first logic if it's a file, and latter if its a directory.

Maybe we should make it explicit?
```suggestion
        # If a directory, assume inner "<dirname>.odc-metadata.yaml"
        if source_level1.is_dir():
            level1_metadata_path = source_level1 / (
                source_level1.name + ".odc-metadata.yaml"
            )
        # Otherwise expect a sibling file with ".odc-metadata.yaml" suffix
        else:
            level1_metadata_path = source_level1.with_suffix(".odc-metadata.yaml")
```

(if that's what you'd expect?)

eodatasets3/wagl.py
```diff
@@ -350,6 +350,20 @@ def val(v: Any):
...
+    else:
+        level1_metadata_path = None
+    return level1_metadata_path
```
jeremyh, line 364: This has slightly strange fallback behaviour if it doesn't exist. We really only want that first logic if it's a file, and latter if its a directory.

Maybe we should make it explicit?
```suggestion
"""Find the expected matching metadata file for the source dataset """
    source_level1 = Path(get_path(wagl_doc, ("source_datasets", "source_level1")))
    if not source_level1.exists():
        return None

    # If a directory, assume "<dirname>.odc-metadata.yaml"
    if source_level1.is_dir():
        return source_level1 / (source_level1.name + ".odc-metadata.yaml")
    # Otherwise it's a sibling file with ".odc-metadata.yaml" suffix
    else:
        return source_level1.with_suffix(".odc-metadata.yaml")
```

(if that's what you'd expect?)

jeremyh: This has slightly strange fallback behaviour if it doesn't exist. We really only want that first logic if it's a file, and latter if its a directory.

Maybe we should make it explicit?
```suggestion
def _get_level1_metadata_path(wagl_doc: Dict) -> Optional[Path]:
    """
    Find the expected matching metadata file for the source dataset
    """
    source_level1 = Path(get_path(wagl_doc, ("source_datasets", "source_level1")))
    if not source_level1.exists():
        return None

    # If a directory, assume "<dirname>.odc-metadata.yaml"
    if source_level1.is_dir():
        return source_level1 / (source_level1.name + ".odc-metadata.yaml")
    # Otherwise it's a sibling file with ".odc-metadata.yaml" suffix
    else:
        return source_level1.with_suffix(".odc-metadata.yaml")
```

(if that's what you'd expect?)

dunkgray:
Looks good to me. I've added your suggestion to the branch

jeremyh (approved):

PR #148, opened 2021-04-22
Fix S2 band 8a name normalisation

jeremyh:
S2 introduces a non-numeric band name in 'band08a', which our normalisation doesn't handle.


dunkgray (approved):
Looks good

PR #150, opened 2021-04-23
Support multi-document yaml files in eo3-validate

jeremyh:
This was overlooked as all the eo3 datasets are currently single-file. But a lot of old products are stored together in a single document.

Includes a test.


dunkgray (approved):
Looks good. Good to see tests.

PR #151, opened 2021-04-28
Added dataset separator to derivative folders

MatthewJA:
If `self.dataset_separator_field` is defined then the field should be used to separate datasets. This prevents clobbering for S2. It's currently ignored for derivatives, so this PR adds it.


jeremyh (approved):
Thanks Matthew.

I'd like to clean up that derivate code split at some point. It's duplicating logic (clearly not perfectly) and has no tests or doc-comments.... so it's not clear to me what it's doing differently deliberately and what it's doing differently accidentally.

PR #159, opened 2021-05-12
Expand Landsat prep script to support USGS level 2

joestasks:
Please review L2 related changes.


eodatasets3/prepare/landsat_l1_prepare.py
```diff
@@ -85,13 +101,16 @@
...
+    "st_b6": "st_b6",
     "6": "swir_1",
     "7": "swir_2",
     "8": "panchromatic",
     "9": "cirrus",
+    "st_b10": "st_b10",
     "10": "lwir_1",
     "11": "lwir_2",
     "quality": "quality",
+    "qa_aerosol": "qa_aerosol",
```
jeremyh: We want a human-readable name for these new bands for consistency with the others. (eg. we use “red”, not “b4”.) 

Might be worth seeing what names USGS use in their own docs? (@lanweiwang ?)

eodatasets3/prepare/landsat_l1_prepare.py
```diff
@@ -362,7 +430,7 @@ def prepare_and_write(
         for section, fields in _COPYABLE_MTL_FIELDS[collection_key]:
             for field in fields:
                 value = mtl_doc[section].get(field)
-                if value is not None:
+                if value is not None and p.properties.get(f"landsat:{field}") is None:
```
jeremyh: What was this extra “and” statement protecting against?

(it wont print a warning about setting a value twice if the value is identical, so if it’s needed there might be something fishy happening?)

jeremyh (commented):
Thanks Joe. It’s looking okay to me.

We’ll need some tests of its products before merge so we can track what its outputting, similar to the tests of [L1](https://github.com/GeoscienceAustralia/eo-datasets/blob/eodatasets3/tests/integration/prepare/test_prepare_landsat_l1.py) and others. Give a shout if you want an overview of how they work.

Note there’s a test failure below, it looks like L1 is failing.

(… and ideally clean-up the print/debug and commented-out statements before merge too 🙂 )

eodatasets3/prepare/landsat_l1_prepare.py
joestasks: This was due to multiple sections having the same property, but with different values because the level2 value, which is given precedence in the new logic if it exists, was then being overwritten by a level1 value. My thinking was that if the precedence logic is correct then the first value for something should be accurate.

eodatasets3/prepare/landsat_l1_prepare.py
joestasks: I'll post this query to Slack for you both...

joestasks (commented):
Am I on the right track with integration test? Have I chosen the correct example as a template and can they all be similar to this one; I just need to add the source data, images and expected YAML?

jeremyh:
> Am I on the right track with integration test? Have I chosen the correct example as a template and can they all be similar to this one; I just need to add the source data, images and expected YAML?

Yep!  It runs the packager with a sample input and checks the output matches expected. You’ll just need to take one of your existing datasets to use as the input, with images scaled down using the gdalmerge command as mentioned before (or anything similar)

tests/integration/conftest.py
```diff
@@ -25,6 +25,18 @@
     Path(__file__).parent / "data" / "LC08_L1TP_090084_20160121_20200907_02_T1"
 )
```
jeremyh: I'd rather keep all of this (the fixtures etc) in `test_prepare_landsat_l1.py`, as they're only used within that file. `conftest.py` is for globals to share across multiple test files and is already quite large.

(Feel free to ditch the extra "expect_*" function complexity too. And possibly inline the expected dictionaries completely into the tests without fixtures. That extra complexity existed because other tests are built with multiple packaging types and had to be constructed multiple times differently. But yours have just has one input and one output per test.)

jeremyh (commented):
Looks good :+1: 

It's mergeable as is, though I'd ideally like to keep a handle on test complexity a bit before it grows out of control.

tests/integration/conftest.py
joestasks: I have tidied the test cases, see what you think.

jeremyh (approved):
:+1:

joestasks:
Will now try to close, not sure if I will have permission to merge.

jeremyh:
Merged manually.

PR #161, opened 2021-05-13
Include constellation field in S2 data

jeremyh:
It and the Stac 'mission' field should be inheritable, too.


ohehir (approved):
I approve. I can't see any problems with these additions.
Aaron

PR #165, opened 2021-05-25
Add test case for expected S2 Sinergise outputs

jeremyh:
... and fix error message on unknown wagl level1 paths
(it was always printed as `None`)


dunkgray (approved):
Excellent.  Yay to more testing.

PR #166, opened 2021-05-31
Add usgs metdata handling

alexgleith:


eodatasets3/assemble.py
```diff
@@ -103,6 +103,7 @@ class DatasetAssembler(EoFields):
         "landsat:collection_number",
         "landsat:landsat_product_id",
         "landsat:landsat_scene_id",
+        "landsat:scene_id",
```
jeremyh, line 106: We discussed this a bit in #136, but `landsat:landsat_scene_id` and `landsat:scene_id` are the same property, so I'd rather not duplicate them. We already have eo3 products using the first field name.

Previously when the Stac standard changed (with fields like platform, instrument) we kept our existing eo3 fields for consistency across eo3 products, and added the translation in the [Stac converter](https://github.com/GeoscienceAustralia/eo-datasets/blob/5529d5e9a69dea641e86ba7df5ddf2afaf59e8c5/eodatasets3/stac.py#L18-L31).  That way eo3 docs could stay consistent and stable, our existing code would continue to work, and the latest Stac changes would still be present on conversion.

Sticking to the original field name also means we wont need the [new logic below](https://github.com/GeoscienceAustralia/eo-datasets/pull/166/files#diff-50b29b9a54eb3d340c8a094efb3ec9c6f142098d9bffddb47728806d655c2446R439-R447), and we wont need different sets of metadata-types to handle the different field name choices across products ... which historically people have trouble with.

eodatasets3/assemble.py
```diff
@@ -219,6 +220,10 @@ def __init__(
             self.names = ComplicatedNamingConventionsDerivatives.for_c3_derivatives(
                 self
             )
+        elif naming_conventions == "deafrica_usgs_c2":
```
jeremyh: Should this just be a "deafrica" naming convention? It's not currently defining anything USGS/Landsat specific.

... otherwise we're going to be adding a separate one for every single deafrica product in the future, right?

eodatasets3/assemble.py
alexgleith: Seeing as the USGS STAC docs use `landsat:scene_id`, I think we should include it: https://landsatlook.usgs.gov/sat-api/collections/landsat-c2l2-sr/items

For future reference, properties looks like this:

```
"properties": {
"collection": "landsat-c2l2-sr",
"eo:gsd": 30,
"platform": "LANDSAT_8",
"view:off_nadir": 0,
"datetime": "2021-05-26T22:14:42.761182Z",
"eo:cloud_cover": 83.38,
"view:sun_azimuth": 33.74414161,
"view:sun_elevation": 14.29418577,
"eo:instrument": [
"OLI",
"TIRS"
],
"landsat:cloud_cover_land": 97.39,
"landsat:wrs_type": "2",
"landsat:wrs_path": "074",
"landsat:wrs_row": "093",
"landsat:scene_id": "LC80740932021146LGN00",
"landsat:collection_category": "T1",
"landsat:collection_number": "02"
},
```

eodatasets3/assemble.py
alexgleith: We might need one for deafrica_s2...

alexgleith: I think this will work for both WO and FC. I guess I'm happy to just call it `deafrica` for now, though. I'll change it.

eodatasets3/assemble.py
jeremyh: We might be talking past each-other a bit :slightly_smiling_face: 

My understanding:

- A document can be "eo3" or "stac" but not both. We explicitly have to convert between them already.
- The stac property name has changed from `landsat:landsat_scene_id` to `landsat:scene_id`   (as your point out: USGS uses that field name now), but eo3 documents are already published with the old name.
- This is the same scenario we faced with the `eo:platform`, `eo:instrument` etc fields.

We have two choices:

1. We can start using the new USGS name within `eo3` products, and the `eo3` world will be inconsistent. We will need extra logic to handle the two possible names in every place that reads them, and different ODC metadata types for different products.
or
2. Or we can keep the existing name in the `eo3` world and have our products be consistent. No extra logic, no extra metadata types. When converted to Stac it will convert to the new USGS name.

In _both_ cases our Stac docs will be consistent with USGS. 

In both cases we'll need to add the conversion to our Stac converter (because of old products).

The only decision is whether we want `eo3` to be consistent with itself, or to break it.

I prefer option 2. Our Stac documents will still all be valid and match USGS. We wont need extra code logic to handle both fields everywhere.

(Note: the list of properties in the code we're comment on are `eo3` properties, not Stac properties)

alexgleith: We could add this field as a special case here: https://github.com/opendatacube/odc-tools/blob/develop/libs/index/odc/index/stac.py#L17

My issue is that we already have 200,000 datasets indexed in DE Africa and we're about to add 800,000 more.

I guess I could re-index them including changing the `landsat:scene_id` to `landsat:landsat_scene_id` but I really think it's better to be consistent with the USGS metadata, in this case, rather than with Australian EO metadata.

jeremyh: Okay, if we've already published datasets with the second name we don't have much choice, we're already using both...

(It wasn't about Australia EO consistency but about ODC system consistency.)

alexgleith:
So you're happy @jeremyh?

eodatasets3/model.py
```diff
@@ -535,12 +540,25 @@ def for_s2_derivatives(cls, dataset: EoFields, uri=DEA_URI_PREFIX):
                 "odc:producer",
                 "odc:product_family",
                 "odc:region_code",
-                "odc:dataset_version",
```
jeremyh, line 538: The derivative naming conventions were added without tests, unfortunately. But I believe they fail without a dataset version because its [used in the folder name](https://github.com/GeoscienceAustralia/eo-datasets/blob/15a199a957998107fa3670fd3d6818f7400712a4/eodatasets3/model.py#L580)

(We have so many variants of naming conventions added by different groups that this is quickly becoming very unwieldly...)

jeremyh (commented):
We're going to have to clean-up these naming conventions variants at some point, so I'd like to make sure we have a test example of each one to avoid breaking it in the future.

Here's a quick test (suitable for `test_assemble.py`). Is the bottom what's expected? Note it swaps "platform" and "family" in the product name compared to other naming conventions.

```python
def test_africa_naming_conventions(tmp_path: Path):
    """
    Minimal fields needed for DEAfrica naming conventions
    """
    with DatasetAssembler(tmp_path, naming_conventions="deafrica") as p:

        # Just the fields listed in required_fields.
        p.producer = "digitalearthafrica.org"
        p.datetime = datetime(1998, 7, 30)

        p.platform = "landsat-7"
        p.region_code = "090081"
        p.product_family = "wo"
        p.processed_now()
        p.dataset_version = '2.3.4'

        dataset_id, metadata_path = p.done()

    metadata_path_offset = metadata_path.relative_to(tmp_path).as_posix()
    assert (
        metadata_path_offset
        == "wo_ls_2/2-3-4/090/081/1998/07/30/wo_ls_2_090081_1998-07-30.odc-metadata.yaml"
    )
```

It failed until I added producer, processing time and dataset version fields. So I think the required fields for the `deafrica` naming conventions should be this:
```python
required_fields=(
                "eo:platform",
                "odc:region_code",
                "odc:producer",
                "odc:dataset_version",
                "odc:product_family",
                "odc:processing_datetime",
            ),
```
(unless the above isn't what's expected)

jeremyh:
Oh wow, with no platform specified at all in that test it's marking it as "ls". I assume we don't want that. It's because the derivate code assumes platform is required.

Do we want to include platform in required fields? Or remove it from the name entirely if not specified?

jeremyh:
I believe the way DEA is going is to set multiple platforms for products like wo and fc.
```
p.platforms == ["landsat-5", "landsat-7"]
```

... and the naming conventions convert that into whatever group name fits, like "ls" or "s2".

alexgleith:
You're right, that was silly of me. Fixed to include `eo:platform`.

Not sure about multiple platforms... Stac SPEC says string, not array: https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#instrument

jeremyh:
> Not sure about multiple platforms... Stac SPEC says string, not array: https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#instrument

Yeah, ODC requires a string too. Some products were created that are derived from multiple, though, and started comma-separating them. Such as [wofs summary](https://explorer.dea.ga.gov.au/products/wofs_annual_summary/datasets/5f6df8d0-6979-456d-aec8-e511aa3d8d74).

It's now supported officially in eodatasets, and is converted/sorted etc automatically:

```
>>> # Set multiple platforms
>>> p.platforms == ["landsat-5", "landsat-7"]

>>> # ... which is still stored as a string:
>>> p.platform
'landsat-5,landsat-7'

>>> # ... `platforms` will always return the platform field as a sequence:
>>> p.platforms
{"landsat-5", "landsat-7"}
```

jeremyh (approved):

Issue #132, opened 2021-03-22
Can't use custom `eo:platform` metadata value


valpesendorfer:
I'm trying to create `eo3` dataset metadata for some MODIS data with the `DatasetAssembler`, and I'd like to include whether the data is from Terra or Aqua. The best fit in the metadata definition I've found is the `eo:platform` property.

But when I try to pass in a platform code, I get a `NotImplementedError`, raised in:

https://github.com/GeoscienceAustralia/eo-datasets/blob/e07a3e87e6d6990ef3a6045784930bbb49455bfd/eodatasets3/model.py#L337

and likely also in

https://github.com/GeoscienceAustralia/eo-datasets/blob/e07a3e87e6d6990ef3a6045784930bbb49455bfd/eodatasets3/model.py#L356

meaning I can't use any platform name that doesn't start with `sentinel-1`, `sentinel-2` or `landsat`. 

Is there any reason the `DatasetAssembler` can't just use the user defined platform code if it doesn't match S1/S2/LS ?

jeremyh:
Thanks for the report @valpesendorfer .

In develop, built-in abbreviations are now extensible.

You can also [set abbreviations manually](https://eodatasets.readthedocs.io/en/latest/#eodatasets3.DatasetPrepare.names):
```python
p.names.platform_abbreviated = "ls"
p.names.instrument_abbreviated = "e"
```

PR #171, opened 2021-06-04
Restructure internals for separate naming, metdata and serialisation apis

### LINKED ISSUE #128: Add support for retrieving a  `DatasetDoc` from `DatasetAssembler`
I'd like to add support for getting a `DatasetDoc` from `DatasetAssembler`. Maybe this could be done with something as simple as:

```python
class DatasetAssembler:
     ...
    def done(..., write_metadata=True): # option to not write metadata to file
        ....
        self.dataset = DatasetDoc(...) # store dataset for user retrieva

```

What do you reckon @jeremyh?

### LINKED ISSUE #154: Add Public API for providing destination paths.
Implement a public API to provide destination paths (currently this function is not exposed publicly, but does exist).

This is required for [Statistician](https://github.com/opendatacube/odc-tools/tree/develop/libs/stats) to start using *eo-datasets* for metadata generation.

@Kirill888

### LINKED ISSUE #155: Allow creation of dataset metadata in memory
Allow creation of dataset metadata in memory, so that a separate tool can handle writing the data itself.

Datacube  [Statistician](https://github.com/opendatacube/odc-tools/tree/develop/libs/stats) handles efficient writing to Object Stores and compressing to GeoTIFF itself, but to avoid duplication and incompatibilities we would like to use eo-datasets for the metadata document generation.

This assumes data/COGs/TIFs are already in place.

This may be a duplicate of #128 , I'm not sure.

@Kirill888

jeremyh:
This is a rather large restructure of how we represent our various naming conventions, docs, metadata and internal structure.

Several inner modules of eodatasets3 are now cleaned-up, public and documented:

### New features

- Ability to create datasets and everything entirely in memory: [example](https://jez-eod3.readthedocs.io/en/latest/#creating-documents-in-memory) 
    - (closes #155?) 
    - (closes #128)
- Ability to do file+name generation by itself (such as to determine if a dataset already exists): [example](https://jez-eod3.readthedocs.io/en/latest/#generating-names-paths-alone) 
    - (closes #154?)
    - Ability to calculate those file paths as URLS (`s3:// ... `), but the writer cannot yet write to remote URLs.
- All <sup>(?)</sup> magic is now optional. Names and properties will only be computed if not set manually. Examples:
    - Set custom abbreviations (among other things) to use in name generation: `p.names.platform_abbreviation = 's2'` ([eg](https://jez-eod3.readthedocs.io/en/latest/#naming-things-yourself))
    - Set paths rather than using naming conventions: `p.names.dataset_folder = ...` ([eg](https://jez-eod3.readthedocs.io/en/latest/#eodatasets3.DatasetPrepare.names))
    - Set a geometry rather than it being calculated: `p.geometry = my_shapely_geometry` ([eg](https://jez-eod3.readthedocs.io/en/latest/#avoiding-geometry-calculation))
    - Provide own pixel/grid information in `note_measurement(grid=...)`
    - Measurement file paths can be specified manually in `write_measurement(...)` (and similar ) ([eg](https://jez-eod3.readthedocs.io/en/latest/#eodatasets3.DatasetAssembler.write_measurement))
    - This also means we *wont need dedicated "naming_conventions"* for every single variation people want to use. They can set their own names if they have their own needs.
- There's now class `DatasetPrepare()`, which is like DatasetAssembler, but for  _only_ making a metadata file ([doc example](https://jez-eod3.readthedocs.io/en/latest/#writing-only-a-metadata-doc)). It avoids the extra packaging complexity. It's a drop-in replacement for `DatasetAssembler()` without the `write_*()` methods.
- Public [serialise](https://jez-eod3.readthedocs.io/en/latest/#reading-writing-yamls) module for reading/writing formatted documents, and related types.

### Repository changes

- `DatasetAssembler` has been split in two. It's simpler, with its metadata logic moved to the super-class `DatasetPrepare`
- Internally, locations are now stored as URLs in anticipation of remote-write support in the future 
    - (to backport features from #62).
- A large reorganisation of how our naming conventions are represented. This unpicks the copy-and-pasted code between them (with minor variations), and their differences are more obvious. (it makes the inconsistency of our naming conventions names rather stark, though :)  
    - (#132)
- Vastly expand the read-the-docs documentation, with more usage examples (and inner doctests, run within Github Actions), fixed broken links (turned on "nitpicky" mode, so they're warnings now).
- Added docs for many of the utility types used by the apis. DatasetDoc, GridSpec, exceptions

### Deprecations

- `DatasetAssembler()` now has a deprecation warning for users producing only a metadata doc. They should change it to `DatasetPrepare()`.
- `add_accessory()` is renamed to `note_accessory()`, with deprecation warning. Identical API, but this matches the name of similar functions.

Possible incompatibilities

- The `dataset_location` parameter is now expected to be the same as the ODC's dataset location. ie. it's usually a metadata path. Previously this was a folder that contained the dataset. I think using the same logic as ODC is a better idea.
- The concept of sub-datasets has been removed (multiple differently-named file sets within one dataset). This added a lot of complexity, was only used by ARD, and niche behaviour like this can now be implemented externally using the naming apis.

:heavy_check_mark: All packaging tests still pass, and new ones added.


omad (approved):
Thanks Jeremy, this looks amazing. However...

It's grown far too large for me to meaningfully review. The documentation and test coverage look really good, so I think we merge this, and start trying to use the new APIs, and if we have any troubles come back and address them then.

PR #173, opened 2021-06-25
Fix allow-absolute, and tell people how to embed locations

### LINKED ISSUE #91: Add DatasetAssembly support for outputting the `location` field
The eo3 format added the ability to include the location of a dataset inside the metadata, and datacube 1.8.2+ will use this value instead of the metadata's own location when it is provided.

eo-datasets can parse this field, and keeps track of locations, but the DatasetAssembler API doesn't currently have an option to write it into the document.

Ideally we want to add this before people start using the old workarounds in new products, such as using absolute paths in bands (which eodatasets currently warns about but does not provide an alternative to).

jeremyh:
People often ask on Slack how to write a metadata document that doesn't sit in the same location as their data.

This [adds a section to the documentation with an example of doing this](https://jez-eod3.readthedocs.io/en/latest/#separating-metadata-from-data).

Also includes related changes:

- fixes `allow_absolute_paths=True`, which evidentially didn't have a test in our repository (none of our products used absolute paths...). The doctest now exercises it.
- Changes the default `dataset_location` to be a folder with ending slash, instead of a metadata file, since this will work with more use cases:

       From: 's3://dea-public-data-dev/collections/ls7_nbar/2014/04/05/ls7_nbar_2014-04-05.odc-metadata.yaml'
         To: 's3://dea-public-data-dev/collections/ls7_nbar/2014/04/05/'
The dataset location was never previously output, so we can change this without breaking backwards compat.
- Closes #91

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,6 +20,16 @@ nitpicky = True
 # Show warnings on page so people want to fix them more.
 keep_warnings = True
 
+doctest_global_setup = """
+import numpy
+import tempfile
+from uuid import UUID
+from datetime import datetime
+from affine import Affine
+from rasterio.crs import CRS
+from pathlib import Path
+"""
+
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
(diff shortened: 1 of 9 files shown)


omad (approved):
Thanks Jeremy!

PR #178, opened 2021-07-05
Use ODC's own doc reader in `eo3-validate`

jeremyh:
This way we can support all the same file extensions (eg. `.yaml.gz`), as well as external URLs:
```
❯ eo3-validate https://explorer.dea.ga.gov.au/products/ga_ls_fc_3.odc-product.yaml
 ```


uchchwhash (approved):
LGTM!

Issue #184, opened 2021-07-12
A small number of datasets are slow to package, reporting self-intersection errors


jeremyh:
Reported by @tebadi 

> Hi Jeremy, I'm running LCCS and a small number of tiles run for a considerable amount of time, something close to 2 hours while all the other tiles finish within 30 minutes.  In all of these tiles I see
Ring Self-intersection at or near point ...  and the CPU activity remains minimal.  see a sample below (check the timestamps).  They all happen when writing level4 data:

```
021-07-12 04:11:13,772] {verify.py:119} INFO - Checksumming PosixPath('/tmp/lccs/.odcdataset-clwyq2l3/ga_ls_landcover_class_cyear_2_1-0-0_au_2013-01-01_baregrad-phy-cat-l4d-au.tif')
[2021-07-12 04:11:14,188] {env.py:433} WARNING - CPLE_NotSupported in driver GTiff does not support creation option COUNT
[2021-07-12 04:11:14,188] {env.py:433} WARNING - CPLE_NotSupported in driver GTiff does not support creation option WIDTH
[2021-07-12 04:11:14,188] {env.py:433} WARNING - CPLE_NotSupported in driver GTiff does not support creation option HEIGHT
[2021-07-12 04:11:14,188] {env.py:433} WARNING - CPLE_NotSupported in driver GTiff does not support creation option CRS
[2021-07-12 04:11:14,188] {env.py:433} WARNING - CPLE_NotSupported in driver GTiff does not support creation option TRANSFORM
[2021-07-12 04:11:14,188] {env.py:433} WARNING - CPLE_NotSupported in driver GTiff does not support creation option DTYPE
[2021-07-12 04:11:14,432] {verify.py:119} INFO - Checksumming PosixPath('/tmp/lccs/.odcdataset-clwyq2l3/ga_ls_landcover_class_cyear_2_1-0-0_au_2013-01-01_level4.tif')
[2021-07-12 04:13:18,660] {geos.py:252} INFO - Ring Self-intersection at or near point 3467 46
[2021-07-12 04:13:18,667] {geos.py:252} INFO - Ring Self-intersection at or near point 3451 65
[2021-07-12 04:13:18,684] {geos.py:252} INFO - Ring Self-intersection at or near point 3477 111
[2021-07-12 04:13:18,684] {geos.py:252} INFO - Ring Self-intersection at or near point 3351 113
[...]

PR #186, opened 2021-07-13
Allow for more vectorisation options

jeremyh:
~**note**: only an experiment for now: we'll see how it affects things~. Ready to merge, I think, as it allows new options but should not change existing default behaviour.

This avoids a common issue that surprises people: observations with small holes (such as specklings of nodata pixels, as seen in #184) can blow out the CPU usage of geometry calculation.

Valid data is allowed to be bigger than the actual image, and none of our use cases benefit from holes, so why not fill them
by default?

(this makes scipy a core dependency, rather than optional)

## Options Added

This PR makes valid-data calculation method an enum.

1. `thorough`: Vectorize the valid pixel mask as-is. (**the default**)
2. `filled`: Fill holes in the valid pixel mask before vectorizing. (**a potential new default**. but I haven't changed it in this PR to avoid potential compatibility issues.)
        (Much faster than vanilla if there’s many small nodata holes, as they cause many tiny polygons that are later removed by geom simplification. But mildly slower when no holes exist: ~1 sec.)
3. `convex_hull`:  (suggested by Imam)
        Take convex-hull of valid pixel mask before vectorizing.
        This is much slower than filled, but will work in cases where you have a lot of internal geometry that aren’t holes. Such as SLC-Off Landsat 7 data.
        Requires extra ‘scikit-image’ dependency.
4. `bounds` (a "do nothing" option)
        Use the image file bounds, ignoring actual pixel values.

### Remaining Questions

Is an enum an okay API? Maybe they'll want parameters in the future? (eg. "number of pixels to dilate by"). But maybe in those cases we should just let them set their own `Callable` to mess with the mask themselves in more advanced ways?

### Todo

- [ ] ~Test on more of the existing products. Could `filled` be used universally instead of `thorough`?~
      - (Our integration tests are all scaled down tiny images, so aren't a good reflection of impact)
      - *Decided to leave potential default-behaviour change for a future PR*
- [x] Maybe add "geometry calc method" as a config enum? (Possible options: filled-holes, normal, full-image-bounds...)
- [x] Mention in docs the valid-data-methods
- [ ] ~If `method==bounds`, the `note_measurement` methods could avoid reading any pixels at all. Currently, this option just avoids the shape calculation.~
   - *Performance optimisation can be left for a future PR*

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -295,7 +295,11 @@ all measurements.
 By default, the assembler will create this geometry by reading the pixels from your
 measurements, and calculate a geometry vector on completion.
 
-If you want to avoid these reads and calculations, you can set the geometry manually::
+This can be configured by setting the :attr:`p.valid_data_method <eodatasets3.DatasetPrepare.valid_data_method>`
+to a different :class:`ValidDataMethod<eodatasets3.ValidDataMethod>` value.
+
+But you may want to avoid these reads and calculations entirely, in which case you can set
+a geometry yourself::
 
     p.geometry = my_shapely_polygon
 
@@ -736,4 +740,5 @@ Misc Types
 
 .. automodule:: eodatasets3
    :members:
+   :member-order: bysource
    :exclude-members: DatasetAssembler, DatasetPrepare, NamingConventions, namer
--- a/eodatasets3/__init__.py
+++ b/eodatasets3/__init__.py
@@ -1,6 +1,6 @@
 from ._version import get_versions
 from .assemble import DatasetAssembler, DatasetPrepare, IfExists, IncompleteDatasetError
-from .images import GridSpec
+from .images import GridSpec, ValidDataMethod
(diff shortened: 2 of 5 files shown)


ArpitKubadia:
Hello,
Has this functionality been added in the latest released version?
For my application, I am using the ValidDataMethod.bounds from this
However the latest official update which adds multiple bands support is also what I need.

Are there any plans to merge these two?
Thanks!!

jeremyh:
> Hello, Has this functionality been added in the latest released version? For my application, 

No sorry. I was hesitant to merge it as changing the default option can be dangerous.

I've just changed this PR to use the existing default, so at least we can merge it as-is safely and allow people to try the new options

alexgleith (approved):

jeremyh:
Released in [version 0.24](https://github.com/opendatacube/eo-datasets/releases/tag/eodatasets3-0.24.0)

PR #188, opened 2021-07-20
Add USGS C2 RMSE fields

alexgleith:
I didn't namespace these. I _could_, but I didn't. Do you think I should?


jeremyh:
They do look like they should be namespaced, but if USGS is using those field names we’re stuck with them, presumably?

alexgleith:
Well... these ones are custom fields that I added manually to the STAC doc. They're lifted out of their MTL.json file. So I can rename them.

Do you think it's worth adding a `landsat:` to the front of them?

jeremyh:
Yep, we did that with the other mtl fields: “landsat:” in front of the given name. So it sounds good to me for consistency.

jeremyh (approved):

PR #189, opened 2021-07-21
Allow skip properties to add_source_dataset

supermarkion:
Add inherit_skip_properties list to add_source_dataset(). 

Some properites should not be inherited directly from source datasets in odc-stats. E.g. gqa:mean_x. 

We add this inherit_skip_properties list to allow the users to skip some properties.


supermarkion:
Give me moment to fix lint.

PR #190, opened 2021-07-21
Feature/odc stats integration

supermarkion:
Add a feature which allows specifying a list of extra property names that shouldn't be inherited when adding source datasets.

This is required for summary datasets. They have a large list of source datasets, and need to not inherit per dataset metadata like GQA, extents and more. E.g. gqa:mean_x.

Add an `inherit_skip_properties` argument to `DatasetPrepare.add_source_dataset()`, of type `Optional[List[str]]`.

### DIFF
--- a/eodatasets3/assemble.py
+++ b/eodatasets3/assemble.py
@@ -553,6 +553,7 @@ class DatasetPrepare(Eo3Interface):
         classifier: Optional[str] = None,
         auto_inherit_properties: bool = False,
         inherit_geometry: bool = False,
+        inherit_skip_properties: Optional[str] = None,
     ):
         """
         Record a source dataset using its metadata document.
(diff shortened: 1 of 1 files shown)


omad (approved):
Looks good to me Sai. I've included the new argument in the docstring and fixed the typespec.

What do you think @jeremyh ?

jeremyh:
That sounds okay to me. I'm mildly iffy about making this a blocklist rather than a allowlist. Say, a new gqa property was added, it would start leaking through to the downstream products, and they'd need to be updated with new block lists. Would it better to just be explicit in which properties we want on the current dataset? I'm not sure.

omad:
> That sounds okay to me. I'm mildly iffy about making this a blocklist rather than a allowlist. Say, a new gqa property was added, it would start leaking through to the downstream products, and they'd need to be updated with new block lists. Would it better to just be explicit in which properties we want on the current dataset? I'm not sure.

That's a good point! It would have to be a new field in datasets, as well as a new field in the current list of fields to copy...

My concern was along similar lines, in that it gets hard to reason about what's going on when you combine inclusion and exclusion lists.

omad:
@supermarkion How would you feel about converting this to be a white list instead of a black list?

supermarkion:
> @supermarkion How would you feel about converting this to be a white list instead of a black list?

I fully agree the allow+skip bundle is bad. 

However, the current allow list is a build-in list in EO3 library. Expose the allow-list will force all the current users to maintain their own column list.

PR #193, opened 2021-07-23
Add default grid to STAC metadata

supermarkion:
Hi, @jeremyh, as we discussed/reviewed this morning. Added the default value to support  STAC Projection (proj) Extension.


eodatasets3/stac.py
```diff
@@ -100,7 +100,7 @@ def _proj_fields(grid: Dict[str, GridDoc], grid_name: str = "default") -> Dict:
     """
     Add fields of the STAC Projection (proj) Extension to a STAC Item
     """
-    grid_doc = grid.get(grid_name)
+    grid_doc = grid.get(grid_name or "default")
```
omad, line 112: @jeremyh Can you please clarify what case this is handling, it's a confusing looking line of code now.

jeremyh:
Did this fix your issue Sai? I was surprised to see that the tests didn't need updating... but looking at them now, the tests do correctly have `proj:transform` on the expected outputs. So what case did we fix?

eodatasets3/stac.py
jeremyh: I'll comment below, as there's maybe more investigation to do...

jeremyh:
To add the backstory for this PR: 

The grid name is optional on EO3 measurements, and if not specified in the document it's assumed to be the 'default' grid:

```
measurements:
  nbar_panchromatic:
    path: ga_ls8c_nbar_3-1-0_088080_2020-05-25_final_band08.tif
    grid: panchromatic
  nbar_red:
    path: ga_ls8c_nbar_3-1-0_088080_2020-05-25_final_band04.tif
  nbar_swir_1:
    path: ga_ls8c_nbar_3-1-0_088080_2020-05-25_final_band06.tif
```
(latter two are 'default' grid)

Sai found that `proj:` information is not being included in his Stac measurements that use the default grid... because this method is getting a `None` `grid_name` instead of `"default"`.

https://github.com/GeoscienceAustralia/eo-datasets/blob/5529d5e9a69dea641e86ba7df5ddf2afaf59e8c5/eodatasets3/stac.py#L103-L108

 But it's strange that it seems to work elsewhere fine? In our tests and other Stac outputs.

supermarkion:
Yes. I just confirmed again. This changes fixed my issue. My issue is:

Expected output:
`"assets": {
        "count_clear": {
            "eo:bands": [
                {
                    "name": "count_clear"
                }
            ],
            "type": "image/tiff; application=geotiff",
            "roles": [
                "data"
            ],
            "href": "ga_ls_wo_summary_x07y23_2000--P1Y_count_clear.tif",
            "proj:shape": [
                3200,
                3200
            ],
            "proj:transform": [
                30,
                0,
                -2016000,
                0,
                -30,
                -3168000,
                0,
                0,
                1
            ]
        }
    }`
Actual output:
`"assets": {
        "count_clear": {
            "eo:bands": [
                {
                    "name": "count_clear"
                }
            ],
            "type": "image/tiff; application=geotiff",
            "roles": [
                "data"
            ],
            "href": "ga_ls_wo_summary_x07y23_2000--P1Y_count_clear.tif",
        }
    }`

As you said, my assests lost their proj relative information.

jeremyh (approved):

PR #195, opened 2021-07-29
Add RAM base thumbnail support

supermarkion:
Hi, @jeremyh and @omad, 

How are you.

As we discussed, I added a numpy-base thumbnail feature. I modified two sections here:

1) moved the applying bit/lookup_table relative codes to a standalone method **filter_singleband_data()**. This new standalone method input and output are numpy.array. So it can be re-used by path-base thumbnail and numpy-base thumbnail.

2) added **create_thumbnail_from_numpy()** and **create_thumbnail_singleband()**. Their input are numpy.array, and output are bytes (the bytes is ready to be dump as thumbnail jpg file). They are using the In-Memory Files from rasterio.io to handle the projection.

I create these changes as the proposal. If we are happy with my API changes, I will add relative unit-test later.

Cheers

Sai

### DIFF
--- a/eodatasets3/images.py
+++ b/eodatasets3/images.py
@@ -958,10 +958,15 @@ class FileWrite:
                 "Please set either bit or lookup_table, you haven't set either of them"
             )
 
-        # the output_data has three layers, as R, G, B
         out_data, stretch = self.filter_singleband_data(input_data, bit, lookup_table)
+
+        if bit:
+            rgb = [out_data, out_data, out_data]
+        else:
+            rgb = out_data
+
         return self.create_thumbnail_from_numpy(
-            rgb=out_data,
+            rgb=rgb,
             static_stretch=stretch,
             input_geobox=input_geobox,
             nodata=nodata,
--- a/eodatasets3/images.py
+++ b/eodatasets3/images.py
@@ -19,7 +19,7 @@ from rasterio import DatasetReader
 from rasterio.coords import BoundingBox
 from rasterio.crs import CRS
 from rasterio.enums import Resampling
-from rasterio.io import DatasetWriter
+from rasterio.io import DatasetWriter, MemoryFile
 from rasterio.shutil import copy as rio_copy
 from rasterio.warp import calculate_default_transform, reproject
 from shapely.geometry.base import CAP_STYLE, JOIN_STYLE, BaseGeometry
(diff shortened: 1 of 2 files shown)


eodatasets3/images.py
```diff
@@ -814,6 +814,79 @@ def create_thumbnail(
...
+        out_scale=10,
+        resampling=Resampling.average,
+        static_stretch: Tuple[int, int] = None,
+        percentile_stretch: Tuple[int, int] = (2, 98),
+        compress_quality: int = 85,
+        input_geobox: GridSpec = None,
+        nodata: int = -999,
+    ):
+        """
+        Generate a numpy which can save as jpg image using the given three arrays as red, green, blue.
```
jeremyh, line 829: Make it slightly clearer why we're duplicating the function
```suggestion
        Generate a thumbnail as numpy arrays.
        
        Unlike the default `create_thumbnail` function, this is done entirely in-memory. It will likely require more
        memory but does not touch the filesystem.
```

jeremyh (commented):
Looks okay, though I haven't run it. I'm slightly uneasy about not testing this within this repo, as I could accidentally break your scripts by accident in the future. But it's not part of the main, documented API so it's not critical.

eodatasets3/images.py
```diff
@@ -878,6 +936,147 @@ def create_thumbnail_singleband(
...
+        # the output_data has three layers, as R, G, B
+        out_data, stretch = self.filter_singleband_data(input_data, bit, lookup_table)
+        return self.create_thumbnail_from_numpy(
+            rgb=out_data,
+            static_stretch=stretch,
+            input_geobox=input_geobox,
+            nodata=nodata,
+        )
+
+    def filter_singleband_data(
```
jeremyh, line 975: I like that you removed the duplication here :+1: 

I'd leave this method as private, though, to keep our public footing small.

```suggestion
    def _filter_singleband_data(
```

eodatasets3/images.py
supermarkion: Thanks a lot for the text.

eodatasets3/images.py
supermarkion: Make sense to me.

omad (approved):
Looks good to me.

PR #203, opened 2021-08-18
Handle NaNs in stac documents

jeremyh:
As reported by @uchchwhash 

We were outputting raw `NaN` values in our stac json document, which is invalid json.

Convert them to strings using ODC's standard method.

Updated the test conversion to include a NaN.

 (... but I'm surprised python's default json module quietly outputs invalid json)

### DIFF
--- a/eodatasets3/scripts/tostac.py
+++ b/eodatasets3/scripts/tostac.py
@@ -10,6 +10,7 @@ from uuid import UUID
 
 import click
 from click import echo, style
+from datacube.utils import jsonify_document
 
 import eodatasets3.stac as eo3stac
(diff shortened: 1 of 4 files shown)


alexgleith:
Apparently it's a bit mixed, in terms of support. https://docs.python.org/3/library/json.html#basic-usage

We perhaps should be writing it as `null`, but I'm worried about serialisation/de-serialisation if we start encoding as strings.

alexgleith:
Actually, seems like Python knows what to do with quoted "NaN", so I'm ok with this!

alexgleith (approved):

uchchwhash (approved):
💯

PR #204, opened 2021-08-18
Udpate STAC Fields

alexgleith:
* Updated from DE Africa use of USGS C2 Landsat


jeremyh:
Are we sure we want to duplicate the epsg/shape/transform fields in the properties? 

The stac converter converts them to eo3 grid+crs fields, so it probably should remove the original stac fields?

(otherwise, how do we choose between them when we have both?

jeremyh (approved):

alexgleith:
According to the Proj extension, shape and transform are supported in properties as well as in assets. They're overwritten by the asset versions if they exist.

Issue #208, opened 2021-09-07
New version of requests-cache breaks eodatasets3


Kirill888:
this one just got release;

https://pypi.org/project/requests-cache/0.8.0/

and it breaks `eodatasets3` 

```
Traceback:
tests/env/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
libs/stats/tests/test_proc.py:2: in <module>
    from odc.stats.proc import TaskRunner, TaskRunnerConfig
libs/stats/odc/stats/proc.py:18: in <module>
    from .io import S3COGSink
libs/stats/odc/stats/io.py:33: in <module>
    from eodatasets3.scripts.tostac import dc_to_stac, json_fallback
tests/env/lib/python3.8/site-packages/eodatasets3/scripts/tostac.py:15: in <module>
    import eodatasets3.stac as eo3stac
tests/env/lib/python3.8/site-packages/eodatasets3/stac.py:20: in <module>
    from requests_cache.core import CachedSession
E   ModuleNotFoundError: No module named 'requests_cache.core'
```

JWCook:
That line just needs to be changed to:
```python
from requests_cache import CachedSession
```

Or pin the requirement `requests_cache~=0.7.4`.

jeremyh:
Thanks!

The fix from #209 is included in release [0.22.2 ](https://github.com/GeoscienceAustralia/eo-datasets/releases/tag/eodatasets3-0.22.2)

PR #209, opened 2021-09-08
Fix import of CachedSession (#208)

Kirill888:
might need to set minimum version for requests_cache with that change, but not sure what the minimal version would be.


JWCook:
That import works as far back as requests-cache 0.3 (from 2013), but I'd recommend at least 0.6 for some important bugfixes.

jeremyh (approved):

jeremyh:
Thanks

PR #212, opened 2021-09-09
Move Dockerfile to the plain gdal image, and publish it

jeremyh:
Mostly following similar conventions to [datacube-alchemist](https://github.com/opendatacube/datacube-alchemist), but we're still freezing dependencies with pip-compile.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Get unstable git tag
+        run: >
+          echo "UNSTABLE_TAG=$(git describe --tags | sed 's/eodatasets3-//')" >> $GITHUB_ENV
+
+      - name: Log the unstable tag
+        run: echo $UNSTABLE_TAG
+
+      - name: Build and Push unstable + latest Docker image tag
+        if: github.event_name != 'release'
+        uses: whoan/docker-build-with-cache-action@v4
```
alexgleith: This is probably not recommended any more.

Check this out. https://github.com/digitalearthafrica/deafrica-tools/blob/main/.github/workflows/build-test-docker.yml

Up to you which way you choose to go!

alexgleith (dismissed):
Approved. There's an option to do simpler caching, but up to you if you want to do it.

uchchwhash (dismissed):

.github/workflows/docker.yml
jeremyh: Ooh, thanks for pointing it out. I was just cargo-culting without reading the docs myself.

I ended up removing caching, as this build step is for publishing images, not for running tests, and with the gdal base image it seems to only take a minute to build anyway.

Also switched it to use the `docker/metadata-action` to extract the version number / tags, as it looks like it has nicer defaults. I'd like to move this repo to use more typical branch+tag naming schemes anyway, so may as well do that now.

PR #218, opened 2021-11-02
Allow reading from multi-band images.

jeremyh:
```python
    p.write_measurement("name_of_band", name_of_band_path, index=2)
```

... and add a known property from existing telemetry.


dunkgray (commented):
looks good

dunkgray (approved):

PR #220, opened 2021-11-09
Fix support for gridless stac conversions

jeremyh:
Fix Stac conversion of datasets that don't have shape+transform information (from #216).

(DEA's older EO1 datasets don't have this information)

This was caught by Explorer's stac-api tests after the recent eodatasets release. I've expanded the test cases here to match.


supermarkion (approved):
LGTM, thanks for this change

PR #222, opened 2021-12-09
Stac 1.0 allows for null geometry

jeremyh:
Related to https://github.com/opendatacube/datacube-explorer/pull/351, 
- Stac 1.0.0 allows null geometry. And so does eo3.


dunkgray (approved):
looks good

PR #225, opened 2022-02-20
s2 clean-up + embedded locations

jeremyh:
For s2 preparation:

- Add a `--provider` option, to only scan for the given provider's type of packages.
- Add an `--embed-location` option, for when users want to place the metadata in a different location to the data
- Rename `sentinel-l1c` to `sentinel-l1` for consistency with other commands.


dunkgray (commented):
Looks good

dunkgray (approved):

PR #226, opened 2022-02-21
Add pystac-native conversion

jeremyh:
Clients that use pystac themselves would rather receive pystac objects, not dicts.

(also: remove the unnecessary http caching. We no longer manually download schemas)


alexgleith (approved):
Fantastic simplification!

PR #227, opened 2022-02-27
Per-field validation. and use embedded locations during measurement validation

jeremyh:
- During validation, check that all defined metadata fields exist in the dataset
- If a location is embedded in the dataset, it should be used when reading measurements. This more closely matches the behavior of ODC. 
    - (this fixes S2 L1 data validation, which requires reading data from `zip://` locations)


dunkgray (approved):
Good stuff.

PR #229, opened 2022-03-08
Split OA from contiguity resolution choice

jeremyh:
... and make only contiguity default to the larger size


dunkgray (approved):
looks good

PR #233, opened 2022-04-01
S2 dataset filtering and tests

jeremyh:
Includes tests for the somewhat-complicated embedded locations too.


dunkgray (approved):
Looks good Jeremy, thanks.

PR #237, opened 2022-04-22
S2 cloudless

uchchwhash:


eodatasets3/wagl.py
```diff
@@ -452,6 +458,31 @@ def for_path(
                 with fmask_doc_path.open("r") as fl:
                     [fmask_doc] = loads_yaml(fl)
 
+                if "sentinel" in wagl_doc["source_datasets"]["platform_id"].lower():
+                    s2cloudless_prob_path = s2cloudless_prob_path or wagl_hdf5.with_name(
+                        f"{granule_name}.prob.s2cloudless.tif"
+                    )
+                    if not s2cloudless_prob_path.exists():
+                        raise ValueError(f"No s2cloudless probability image found at {s2cloudless_prob_path}")
```
jeremyh, line 466: Just to confirm -- do we want s2cloudless to be mandatory? This will refuse to package any older (pre-s2-cloudless) wagl outputs. Will wagl *always* include it from now on?

jeremyh (approved):
Looks good

eodatasets3/wagl.py
uchchwhash: That's what I think at the moment. Maybe I am wrong. Will bug you again if I turn out to be wrong. 😄

PR #239, opened 2022-04-26
Add support for accessory files hosted on s3.

tebadi:


eodatasets3/verify.py
```diff
@@ -91,6 +109,21 @@ def add_file(self, file_path):
         :type file_path: Path
         :rtype: None
         """
+
+        if "s3" in str(file_path):
+            s3client = boto3.client('s3', region_name='us-east-1')
+            bucket, key = get_bucket_key(filename)
```
jeremyh: This is causing the test failure --- should be `file_path`, I think

eodatasets3/verify.py
```diff
@@ -39,8 +50,15 @@ def calculate_file_hash(filename, hash_fn=hashlib.sha1, block_size=4096):
     :return: String of hex characters.
     :rtype: str
     """
-    with Path(filename).open("rb") as f:
+    if "s3" in str(filename):
```
jeremyh: This is a little dangerous as it could be triggered by the characters 's3' anywhere in any file path.

For now, `.startswith("s3://")` would suffice as a safer option. Or most "correctly": use `is_uri()` to check if it's a URI, and `urlparse()` (as above, in get_bucket_name) to get the scheme, to see if it's `s3`.

eodatasets3/verify.py
```diff
@@ -91,6 +109,21 @@ def add_file(self, file_path):
         :type file_path: Path
         :rtype: None
         """
+
+        if "s3" in str(file_path):
```
jeremyh: (same as above comment)

jeremyh (commented):
Thanks Toktam.

Just a few minor things (and the tests should be able to pass on the current version).

Not required for this PR, but worth noting:
- Longer-term I'd like to make it support all URLs (like `http`) to make it consistent with our other URL support. But supporting S3 on it's own is a step forward.
- The S3 reading loads the entire file into memory, which could surprise people -- we might want to keep an eye on whether to stream/chunk it in the future.

(edit:)
- And we may want to look at how s3 region_name is chosen -- whether there's a consistent way to specify it (environment variables? config options?) rather than hard-code one.

eodatasets3/verify.py
tebadi: You're right -I changed this to file_path.

eodatasets3/verify.py
tebadi: Done.

eodatasets3/verify.py
tebadi: Great suggestion, done.

tebadi:
> * reading loads the entire file into memory,

Good points here!  I've already changed the AWS region to be an environment variable.

jeremyh:
Thanks Toktam. The current tests work now. You might want to recheck the `s3` support, as I'm pretty sure some changes broke it.

The CI build is failing because of a missing import and some minor things -- I recommend [running pre-commit](https://github.com/opendatacube/eo-datasets#pre-commit-setup) to see those errors, and to auto-format.

eg:
```
pip install pre-commit
pre-commit install
pre-commit run -a
```

tebadi:
> Thanks Toktam. The current tests work now. You might want to recheck the `s3` support, as I'm pretty sure some changes broke it.
> 
> The CI build is failing because of a missing import and some minor things -- I recommend [running pre-commit](https://github.com/opendatacube/eo-datasets#pre-commit-setup) to see those errors, and to auto-format.
> 
> eg:
> 
> ```
> pip install pre-commit
> pre-commit install
> pre-commit run -a
> ```

Thanks Jeremy.  I fixed the pending issue and it's looking good:

image

All the workflows are also passing.

eodatasets3/verify.py
```diff
@@ -39,8 +58,15 @@ def calculate_file_hash(filename, hash_fn=hashlib.sha1, block_size=4096):
     :return: String of hex characters.
     :rtype: str
     """
-    with Path(filename).open("rb") as f:
+    if is_uri(str(filename)):
+        bucket, key = get_bucket_key(filename)
+        s3client = boto3.client('s3', region_name='us-east-1')
```
jeremyh: We should take the new region environment-variable here too

tebadi: I missed this one last time, done.

jeremyh (approved):
Nice — thanks Toktam! 👍

PR #240, opened 2022-04-28
ESA's next collection will be "collection 1"

jeremyh:
... as raised by Duncan. So we should call their current level1 data Collection 0.

This will require updates to our products to match.


dunkgray (commented):
Looks good.  Thanks Jeremy.

dunkgray (approved):

PR #241, opened 2022-04-29
Add multiprocessing, error handling for S2 L1 Prepare

jeremyh:
Improve support for batch preparation

- Allow individual failures to happen without killing the whole process.
- Allow multiple workers to be spawned by the preparer directly. (We were doing this with gnu parallel but pay the high startup time in environments like NCI, and it makes some things like reporting difficult)
- Allow input paths to be specified from a file


dunkgray (approved):
Looks good

PR #242, opened 2022-04-29
Run linting without a volume

jeremyh:
Fixes the linting on Github CI


dunkgray (approved):
I was too fast in doing the last commit!

PR #243, opened 2022-05-04
Quality improvements to folder scanning, filtering, logging and reliability of s2 prepare

jeremyh:
Some minor quality improvements from running it at NCI, and from writing a few docs.

- Improved folder scanning (many fs operations are very slow at on NCI's large filesystems)
- Show interactive progress when running, and improve logging output when run from a non-terminal.
- Add tests for more folder naming edge cases in our collections.
- Add simple example of filtering to the readme.
- Add a `--dry-run` option to see what will be found/filtered.
- Allow filtering to be done on folders, not just file paths.
- Improve path normalisation. We want to maintain the paths the user specified, not convert symlinks.
- Run file scanning in parallel with preparation workers
- Use S2 Tile metadata over product metadata when they both exist


uchchwhash (approved):
Looks really good to me!

PR #244, opened 2022-05-09
Add an --index option to s2 prepare

jeremyh:
The `--index` option allows it to index newly-created datasets immediately into a datacube. It will use the default datacube, but the standard `-E` and `-C` options (and environment variables) are available for specifying a datacube config/environment.

(... and update the readme)

```
❯ eo3-prepare  sentinel-l1 --index  \
    ~/dea/test-data/L1C/2022/2022-03/25S125E-30S130E/S2B_MSIL1C_20210719T010729_N0301_R045_T53LQC_20210719T021248.zip 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,12 +215,11 @@ can be as follows:
 module use -a /g/data/v10/private/modules/modulefiles /g/data/v10/public/modules/modulefiles
 module load eodatasets3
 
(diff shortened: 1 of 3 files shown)


dunkgray (approved):
Looks good

PR #245, opened 2022-05-11
S3 scanning region list should be a whitelist

jeremyh:
- Update the tests
- ... and give the parameter a name that reflects that clearly (`--only-regions-in-file`).


dunkgray (approved):
Nice

PR #246, opened 2022-05-11
Allow products+types to customise validation

jeremyh:
This is less fragile than turning off validation manually from the command line: each product can say what exceptions it _expects_ in its own datasets.

For example, the product definition could be updated to say that geometry is optional on its datasets:

```
default_allowances:
    require_geometry: False
```

... and then the validator will not throw an error for it.

This is driven by the desire to make the DEA product config more manageable -- we can update products with their validation settings to fix the current build.


dunkgray (approved):
This one was harder to follow.

PR #247, opened 2022-05-17
Fix custom ODC config, while allowing it to be optional

jeremyh:
We need to use our own `@pass_config`, as the built-in ODC is too strict.


dunkgray (approved):
Thanks

PR #249, opened 2022-05-18
Expand docs for validator (eg. per-product config)

jeremyh:


dunkgray (approved):
looks like black had a few suggestions

PR #250, opened 2022-05-18
Do subfolder calculation *after* expanding folders

jeremyh:
- Fixes lack of sub-folders (and error about no relative path found) when directory scanning is used in some cases.
- Normalise all folder paths consistently.
- Includes tests.


dunkgray (approved):
At some stage the readme should also be updated so the user knows how this works.

PR #251, opened 2022-05-22
Fix indexing of datasets with embedded locations

jeremyh:
Explicitly always give ODC a location for indexed datasets.

(When the location is inside the document, rather than external via filesystem, the normal indexing api was not recognising them.)


dunkgray (approved):
Thanks

jeremyh:
Thanks @dunkgray  -- fix has been deployed in 0.28.1

PR #253, opened 2022-06-03
Pre-'16 and multi-granule S2 prep support

jeremyh:
- Update metadata extraction to work with the older naming schemes
- Support multi-granule datasets. Output multiple metadata files with the granule id included in file name.
- Use logging rather than a quiet progress bar, as users almost always want to see what's actually being processed for a long running script like this.
- Include test that runs an older dataset with multiple granules.


ohehir (approved):

dunkgray (approved):
looks good

PR #254, opened 2022-06-07
Make granule-id in filename configurable

jeremyh:


ohehir (approved):
Approved

dunkgray (approved):
Thanks.  I was wondering what was needed for this.

PR #255, opened 2022-06-09
Add a cheaper way to skip datasets

jeremyh:
It can be really expensive to check all inner granules of every dataset to know what has been processed already. 

This PR changes the default behaviour to skip datasets immediately if _any_ outputs exist at all, using their filename. 

The downside of this is that a multi-granule dataset that whose packaging was interrupted may have some granules missing from the outputs, and this will skip it rather than check them all.

I've added the flag `--thoroughly-check-existing` to make it (run slower and) actually check all inner granules are processed.

Example:
```
eo3-prepare sentinel-l1 \
    --throughly-check-existing  \
    dev/S2A_MSIL1C_20201031T004711_N0209_R102_T53JQJ_20201031T022859.zip \
    dev/S2A_OPER_PRD_MSIL1C_PDMC_20161213T162432_R088_V20151007T012016_20151007T012016.zip 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,10 @@ Options:
                                   for multi-granule files). Beware that multi-
                                   granule datasets without a granule id in the
                                   filename will overwrite each-other
+  --throughly-check-existing / --cheaply-check-existing
+                                  Should we open every dataset to check if
+                                  *all* inner granules have been produced?
+                                  Default: false.
   --provider [sinergise.com|esa.int]
                                   Restrict scanning to only packages of the
                                   given provider. (ESA assumes a zip file,
--- a/eodatasets3/prepare/sentinel_l1_prepare.py
+++ b/eodatasets3/prepare/sentinel_l1_prepare.py
@@ -554,6 +554,13 @@ class InputDataset:
     "Beware that multi-granule datasets without a granule id in the filename will overwrite each-other",
 )
 @click.option(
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -242,6 +242,10 @@ Options:
                                   for multi-granule files). Beware that multi-
                                   granule datasets without a granule id in the
                                   filename will overwrite each-other
+  --throughly-check-existing / --cheaply-check-existing
+                                  Should we open every dataset to check if
+                                  *all* inner granules have been produced?
+                                  Default: false.
```
dunkgray, line 248: *all* inner granules have been produced?
Needed for pre-2010(?) multi-granule data.
                                  Default: false.

dunkgray (approved):
Nice! Thanks. line 810, 814 etc comments are great

README.md
jeremyh: It's not _strictly_ needed for pre-2016 data, because it will still run through all of the inner multi-granules when processing those datasets

This only affects the skipping of datasets when scanning: It wont check whether _all_ of the granules individually are done before skipping the package; If one is visible, it assumes they're all completed.

dunkgray:
How about
not needed for post-2016 data?
or
Not needed for non multi-granule data?

Just need a way to practically know when this should be used.

PR #257, opened 2022-06-21
create eo3 validate action

### LINKED ISSUE #256: create eo3-validate a dockerised github action
instruction is available here: https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

pindge:
Fixes #256 

based on instruction here https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,17 @@ But all are optional for basic checks
 	  --help                          Show this message and exit.
 
 
+### Eo3-validate docker github action
+#### Inputs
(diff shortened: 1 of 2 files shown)


omad (approved):
Thanks Pin.

This could definitely be useful!

I've got some concerns around maintainability, and how we're building Docker images, but not enough to put off merging this and seeing if it gains traction.

PR #259, opened 2022-07-04
Improve more edge cases of older S2 L1 datasets

jeremyh:
- Add region mapping for older s2 datasets without regions (use the area to compare to a table of previously-known areas)
- Support for a few metadata quirks, like missing resolutions
- Includes a cut-down test dataset


dunkgray (commented):
looks good, thanks Jeremy

dunkgray (approved):
annnd do the approval too

PR #263, opened 2022-07-11
LS9 L1 preparation

jeremyh:
- [x] Add test case with a cut-down dataset.
- [x] Get test case to pass
- [ ] ~Update to match the newer S2 L1 preparation tools~ do in future PR


jeremyh:
Example metadata for @dunkgray's test datasets are here: `/g/data/v10/agdc/jez/c3/ls9`

dunkgray (approved):
looks good

PR #265, opened 2022-08-04
Handle S2 L1 files with extra QA data

jeremyh:
These extra qa xml files have very similar filenames to the normal metadata filenames, so eod3 baulked at finding multiple.

Added this example to our test data.

I've also moved the s2 packager to structlog to match other modules -- the test error messages were fairly useless before.


dunkgray (approved):
Thanks.  Plus that was a good idea to fix the logs.

PR #266, opened 2022-08-05
Read S2 region codes from tile info, not datastrip info

jeremyh:
The datastrip identifier is different on older datasets, and does not contact a regoin code. It also differs on multigranule datasets.

So we should always read region_code from indivdiual tiles, not the datastrip.

- Adds a test of the full metadata for multi-granule datasets (which contained the error).
- Issue reported by Duncan.


dunkgray (approved):
Thanks! It's great to see a test too.

PR #267, opened 2022-08-09
USGS C2 Landsat should include all accessories

jeremyh:
I noticed these were missing in our LS9 data... and fixing it showed we weren't including them in LS 5,7,8 either, despite them being in the package.

This updates test data for all platforms to expect the extra bands in USGS C2 products.


uchchwhash (approved):
LGTM! (github showed me the .jpg! amazing)

PR #269, opened 2022-09-19
inherit setuid/setgid permissions in output folders

jeremyh:
If a folder has the setuid/setgid permissions bits set, then our own subfolders inside them should have it too.

The lack of these bits has caused permissions issues on NCI, where files must be set to the correct group differnetly for each storage location.


dunkgray (approved):
Ha ha, needed that extra 7.

PR #273, opened 2022-10-16
Add filtering test case for older S2 datasets

jeremyh:
.... and improve test error messages for wagl failures.


uchchwhash (approved):
Looks good to me!

PR #274, opened 2022-10-20
Improve stability of Docker container, bump dependencies

jeremyh:
We freeze most python dependencies in Docker, but not the build libraries. They've gotten out of sync and dependency-update no longer works.

- Freeze the build dependencies to prevent this again.
- Ensure `python-gdal` matches base system `gdal`
- Make the `dependency-update` command always update things. It incorrectly used cached versions.
- Bump all dependencies.
- Update tests that were too specific to pyproj versions


omad (approved):
Looks reasonable Jeremy.

I feel like maintaining all these requirements and environment combinations are too complex.

Is it actually worth maintaining a separate Docker Image here?

PR #275, opened 2022-10-24
Select datastrip files for the current granule

jeremyh:
Add code to decide between multiple datastrip files.

Some S2 files we're seeing have multiple datastrips in them (not just multiple granules).

The new code will read all datastrip metadata files to find which one contains the current tile (granule), and only use it.


uchchwhash (approved):
LGTM! Thanks @jeremyh

PR #278, opened 2022-10-31
Limit granule search to granules with imagery

jeremyh:
We've found some S2 datasets which have "null" granule folders. They contain no imagery files. 

To avoid this, we're limiting the granule-listing to only include granule names containing imagery.
    
This also includes a test: I've expanded a test dataset to contain a similar "null" folder.


dunkgray (approved):
Cool. That's great that there's a test, thanks.

PR #287, opened 2022-12-14
Faster eo3 conversion

Ariana-B:
Make remove converter initialisation out of `from_doc` to make it faster, and remove `normalise_properties` param as it is not being used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ import versioneer
 
 HERE = pathlib.Path(__file__).parent.resolve()
 
-README = (HERE / "README.md").read_text()
+README = (HERE / "README.md").read_text(encoding='utf-8')
 
 
 tests_require = [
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -71,6 +71,7 @@
         "ciso8601",
         "click",
         "defusedxml",
+        "h5py",
```
omad, line 79: This is already covered up able as an extra dependency for "all" and "wagl", so shouldn't be necessary here.

omad (changes requested):
Looks good!

I search a bit further and couldn't find anything using the `normalise_properties` argument in other repos either. However, we need some sort of a changelog, to record for the next release that the API has changed.

Can you please do a rebase/force-push to remove these two extra files in `dist/` and we can merge.

setup.py
Ariana-B: This was a change I previously made as part of the conda-forge deployment; without it, their pipeline failed

omad: Okay, thanks, that sounds fair enough.

Is an indicator that we need to be careful about what/when tests are run more than anything I think.

omad (approved):
TY!

PR #296, opened 2023-02-24
Prepared S2 L1 filters that fail indexing should be cleaned up

jeremyh:
We shouldn't leave files around if indexing fails, as this prevents retry of the dataset.

(reported by Duncan)


dunkgray (approved):
looks good.  Have you tested that it works?

jeremyh:
Yep, I've tested variations locally: when overwriting and not overwriting files

dunkgray:
Cool, thanks

PR #299, opened 2023-03-28
add platform for derived as folder part

### LINKED ISSUE #298: destination_folder only supports baseline products and not derived products
destination folder path example works fine for baseline products,  when it comes to derivated products that are processed based on baseline products, it isn't flexible enough to avoid conflict.
```
        Example: ``"ga_ls8c_ard_3/092/084/2016/06/28"``
```

i.e. when both `ls7` and `ls8` passes on the same region on the same day, both tiles will produce the same `region/time` path and the first processed dataset will be lost when the second dataset is processed for the same region and day.

**Proposed solution**: to add platform as part of the path

https://github.com/opendatacube/eo-datasets/blob/develop/eodatasets3/names.py#L293

pindge:
Fixes: #298 

addes `platform abbrev` to end of the dataset folder path if its c3 derivative product.


jeremyh:
As is discussed on Teams, the approval for this as a change is going through other channels, as eo-datasets is just following the upstream "naming conventions" documents that were made.

If it is approved, though, we should probably make this a different `naming_conventions="blah"` variant, rather than modfiying the existing one and breaking backwards compatibility. Products can then choose the move to the new one rather than have datasets start appearing in unexpected places.

PR #304, opened 2023-05-08
Strip core dependencies

SpacemanPaul:
Strip out all dependencies on datacube-core.

Had to copy across some utils code (and switch to odc-geo).  The copied code is all marked `# TODO CORE` so core can be refactored to use these implementations, and otherwise cleanup.

Still a lot of higher level functionality to strip out, so some of the duplicated code from core may be able to be removed later.


SpacemanPaul:
Damn did it again.

PR #313, opened 2023-12-04
Update ls4-7 band aliases

Ariana-B:
Recently downloaded ls7 l1 c2 scenes crash Landsat packaging during the yaml generation phase due to updated band mappings in new collection. Add missing band aliases to mapping.
Also update github actions in docker workflow, and update pystac and jsonschema versions.


jeremyh:
Thanks for looking into this Ariana!

It looks like the test failures are unrelated to the change. A dependency may have broken:

```
self = <pystac.validation.stac_validator.JsonSchemaSTACValidator object at 0x7fd31a235f40>
schema_uri_map = None

    def __init__(self, schema_uri_map: Optional[SchemaUriMap] = None) -> None:
        if not HAS_JSONSCHEMA:
>           raise ImportError("Cannot instantiate, requires jsonschema package")
E           ImportError: Cannot instantiate, requires jsonschema package
```

dunkgray (dismissed):
looks good

Ariana-B:
@jeremyh tests are passing now

jeremyh (approved):
Looks good — nice work

PR #315, opened 2023-12-15
Contiguity fix

geoscience-aman:
**Background**

Sentinel-2 Level 1C data has a quality issue - “4.5 Striping due to lost source packets (no. 10)”, which is highlighted in ESA's Data Quality Report: 

https://sentinels.copernicus.eu/documents/247904/685211/Sentinel-2_L1C_Data_Quality_Report.pdf/6ad66f15-48ca-4e65-b304-59ef00b7f0e0?version=5.0&t=1625665123197

For example, the black squares in the scene below are missing packets. They appear as low values, generally below a `10,000 X` scaled surface reflectance of ~50. Importantly, these low values fall through the contiguity logic, which asks if the `data > 0` across all bands for a given pixel, but are not usable for any practical purpose, as far as I'm aware.

drawing

**This PR**

This PR adds the functionality to check whether a scene has missing packets and if so, to mark the affected pixels as `nodata (-999)` using ESA's missing packet mask. Based on this fix, the contiguity layer should automatically be updated to mark the missing packet pixels as non-contiguous. 

Additionally, the contiguity layer creation has been updated to include all Sentinel-2 bands, as originally intended.

**Testing**
For the following scenes, `wagl.package()` with default parameters has successfully run to completion. The band images and contiguity layer has been visually inspected to ensure the contiguity logic is being followed as expected.

- [x] Original scene above that was identified as problematic (correctly masked using `.gml` vector mask). https://nbviewer.org/gist/geoscience-aman/9576576892651d45060a736a5136159a
- [x] Post January 2022 scene, when ESA changed the format of their mask from `.gml` to `.jp2`. (https://sentinel.esa.int/web/sentinel/technical-guides/sentinel-2-msi/level-1c/masks)
- [x] Landsat scene (unaffected as expected).
- [x] Multi-granule scene.
- [x] Couple extra random Sentinel-2 scenes.


eodatasets3/wagl.py
```diff
@@ -264,29 +473,39 @@ def _create_contiguity(
...
-                    contiguity = numpy.ones((ds.height, ds.width), dtype="uint8")
-                    geobox = GridSpec.from_rio(ds)
-                elif ds.shape != contiguity.shape:
-                    raise NotImplementedError(
-                        "Contiguity from measurements of different shape"
-                    )
-
-                for band in ds.indexes:
-                    contiguity &= ds.read(band) > 0
+                out_shape = ds.shape
```
jeremyh, line 430: If I'm reading this correctly, this isn't guaranteed to be the correct resolution, as we're iterating _all_ measurements, not just the higher resolution one?

jeremyh: Ah, nevermind. It's skipping the non-res ones above.

eodatasets3/wagl.py
```diff
@@ -121,6 +135,161 @@ def _unpack_products(
...
+    if platform_bool:
+        xml_bool = False if missing_packets_xml(granule) is None else True
+    else:
+        return False
+
+    # Mask only if all are True
+    return True if all([band_bool, platform_bool, xml_bool]) else False
+
+
+def mask_h5(dataset, granule, children):
```
jeremyh: This is low priority, but I'm not a huge fan of passing "children" here as a few stray xml element objects. (this means every function needs knowledge of the xml structure instead of just the function that parses it).

jeremyh (commented):
This looks good! Nice work breaking it into clean, logical functions.

We should add a test case (especially so that something runs the xml logic), and once #316 is merged we can verify that the build passes.

We should still:
1) Hand-verify that the masking looks correct on a dataset (is the nearest-resampling giving correct behaviour?)
2) Allow a fall-back for when the Level 1 zip path (read from the HDF5) no longer exists. Currently people can specify a L1 metadata path but not zip path. (but this is not urgent for this PR)

eodatasets3/wagl.py
```diff
@@ -121,6 +135,161 @@ def _unpack_products(
                     )
 
 
+def missing_packets_xml(granule):
+    """
+    Return children xml document if missing packets and None if not
+    """
+    l1_zip_path = Path(granule.wagl_metadata["source_datasets"]["source_level1"])
+    xml_file_path = Path(
+        granule.source_level1_metadata.accessories["metadata:s2_tile"].path
```
jeremyh, line 149: This is great, I forgot that the metadata gave us the tile path. So it should handle all the various package layouts of S2 automatically, including multi-granule.

geoscience-aman:
Thanks for the feedback @jeremyh and @uchchwhash!

- "Hand-verification" here: https://nbviewer.org/gist/geoscience-aman/9576576892651d45060a736a5136159a 
- Have removed passing xml elements and implemented other feedback.
- Level 1 zip path fallback can be a separate PR.
- May need your help for adding the xml logic test case @jeremyh.

eodatasets3/wagl.py
```diff
@@ -121,10 +133,147 @@ def _unpack_products(
...
+        "8": "7",
+        "8A": "8",
+        "9": "9",
+        "10": "10",
+        "11": "11",
+        "12": "12",
+    }
+    esa_band = band_mapping[dataset.attrs["band_id"]]
+    type, path = mtd_dict[esa_band]
+    mask_string_path = f"{l1_zip_path.name.replace('.zip', '.SAFE')}/{path}"
```
jeremyh: I wasn't sure about this assumption (the inner folder always being named the same as the zip, with a different suffix), but it looks like it's correct for our data:

```
sqlite> select count(*), comparison_result from (
(x1...> SELECT 
(x1...>     files.id,
(x1...>     files.file_path,
(x1...>     granules.offset,
(x1...>     CASE 
(x1...>         WHEN SUBSTR(files.file_path, LENGTH(files.file_path) - LENGTH(SUBSTR(files.file_path, INSTR(files.file_path, 'S2'))) + 1, LENGTH(SUBSTR(files.file_path, INSTR(files.file_path, 'S2'))) - 4) || '.SAFE' = 
(x1...>              SUBSTR(granules.offset, 1, INSTR(granules.offset, '/') - 1)
(x1...>         THEN 'Match'
(x1...>         ELSE 'No Match'
(x1...>     END AS comparison_result
(x1...> FROM 
(x1...>     files
(x1...> JOIN 
(x1...>     granules ON files.id = granules.file_id
(x1...> ) group by 2;
916546|Match
```

eodatasets3/wagl.py
```diff
@@ -121,10 +133,147 @@ def _unpack_products(
...
+    except IndexError:
+        return None
+
+    if msi_data_percentage <= 0.0:
+        return None
+
+    children = root[2][1]
+    return {
+        # Append metadata bandId as key and mask type and path as tuple vals to dict
+        e.attrib["bandId"]: (e.attrib["type"], e.text)
```
jeremyh, line 181: I've found in the scan that some of the mask paths (`e.text` here) don't contain a folder prefix -- they're just the file name of the mask. So I had to add logic for both:

This is what that the scanner does to find it:
```
        for mask_xml_tag in doc_root.findall(".//MASK_FILENAME"):
            mask_file_offset = join(product_root, mask_xml_tag.text)
            # Some scenes seem to just have the file name. Assume it's in the QI_DATA folder.
            if mask_file_offset not in zf.namelist():
                mask_file_offset = join(dirname(tile_md), "QI_DATA", mask_xml_tag.text)
```

The product_root variable is calculated earlier (using the location of the metadata file to determine the folder name inside the zip).

eodatasets3/wagl.py
```diff
@@ -121,10 +133,147 @@ def _unpack_products(
...
+    # Parse tile metadata xml file
+    with zipfile.ZipFile(l1_zip_path, "r") as z:
+        return missing_packets_from_xml(z.read(xml_file_path.as_posix()))
+
+
+def missing_packets_from_xml(xml: bytes):
+    root = Et.fromstring(xml, forbid_dtd=True)
+
+    try:
+        msi_data_percentage = float(root[2][0][1].text)
```
jeremyh: `root[2][0][1]` is too fragile, as documents can differ in their tags and we're assuming here a very specific order of them. It's also not very clear what it's trying to select.

`root.find('.//DEGRADED_MSI_DATA_PERCENTAGE')` is an alternative that will find it by name more explicitly.

... but I'll make a separate PR to update which xml fields we use, because it looks from the scan that the msi percentage wont be adequate.

eodatasets3/wagl.py
```diff
@@ -121,10 +133,147 @@ def _unpack_products(
...
+    except IndexError:
+        return None
+
+    if msi_data_percentage <= 0.0:
+        return None
+
+    children = root[2][1]
```
jeremyh: Similar fragility (depends too much on document structure).

See my code snippet in other comment for an alternative.

eodatasets3/wagl.py
```diff
@@ -121,10 +133,147 @@ def _unpack_products(
...
+    Mask a hdf5 dataset using the gml files provided in the path.
+    """
+    data_array = dataset[:] if hasattr(dataset, "chunks") else dataset
+
+    # Open the .jp2 file
+    sp = f"zip+file://{l1_zip_file.as_posix()}!/{mask_string_path}"
+
+    with rasterio.open(sp) as source_ds:
+        source_layer1 = source_ds.read(3)
+        source_layer2 = source_ds.read(4)
```
jeremyh: Perhaps add a source code comment for what it's trying to do (what 3, 4 are)

eodatasets3/wagl.py
uchchwhash: Do you have a recommendation for some alternative logic that could be more robust?

eodatasets3/wagl.py
geoscience-aman: Do you want me to make your suggested changes, you merge this PR and then you submit a separate PR to update the xml fields?

uchchwhash (approved):

jeremyh (approved):
👍 looks good

eodatasets3/wagl.py
jeremyh: Resolving, fixed in the recent commit

PR #321, opened 2023-12-20
Update S2 L1 tests for masking

jeremyh:
- Expand the test datasets, as the packager now needs to read L1 data files for masks, not just the metadata.
- Support for Sinergise package layouts, to fix the Sinergise tests.
- Skip masks that are too small have data (fiona doesn't like the gml files ESA provides that are null)


geoscience-aman (approved):
Have not tested. Looks okay. Thanks @jeremyh!

PR #328, opened 2024-02-12
Fix deprecated `crs.crs_str`

robbibt:
Currently the `dataset_assembler.write_measurements_odc_xarray` method products many deprecation warnings due to the use of `crs.crs_str`:
image

This PR updates the offending line to `str(crs)` instead, as per deprecation warning.


jeremyh (approved):
Thanks Robbi!

(the linting failure seems be in the dev branch, not this PR, so we can fix it separately)

PR #330, opened 2024-02-16
Use existing nodata attributes if they exist on `xarray.DataArrays`

robbibt:
Currently the `write_measurements_odc_xarray()` method requires you to set a single data value to all `xr.DataArrays` in the input `xr.Dataset`. However, individual variables in a dataset often have their own nodata values, as specified in their ODC `.nodata` attribute. 

This PR updates `write_measurements_odc_xarray()` to use the nodata value in the `.nodata` attribute if it exists (and if no custom nodata value is manually provided via the `nodata` param).

Have tested this locally and it works well.


jeremyh (approved):

jeremyh:
Thanks Robbi

PR #334, opened 2024-02-28
Add intertidal properties to known-properties

jeremyh:
(Also add some missing docs, and fix a config warning in the new version of ruff)


robbibt (approved):
This looks great, thanks Jeremy!

PR #335, opened 2024-03-07
Include DEA's new offshore territories tag

jeremyh:
... and add a test case of a modern wagl output.

The added test data found a bug unrelated to the offshore territories: when the first measurement has no explicit CRS it fails with a reference error. 
- (Bands don't all need to declare the CRS explicitly, because they should all be the same CRS. But this failed when it was the first band that was not explicit.)
 

I've also unwisely included a few housekeeping commits here. 
- Enabled DTDs again fro the Modis packager. This was tightened some time ago for our securtiy linter, but the production datasets from the US require them. The processor was only recently upgraded to a (much) newer version of eodatasets so this was never picked up.
-Pre-commit bump


jeremyh:
(The Docker Push action is failing for an unrelated reason — missing dockerhub credentials. I'm not an repo admin so can't view/setup credentials https://github.com/opendatacube/eo-datasets/issues/336)

uchchwhash (dismissed):
Looks good to me!

uchchwhash (approved):

PR #341, opened 2024-04-04
Update BRDF parser for metadata change on the 11th

jeremyh:
The documents USGS provides have changed format on the 11th, causing failures in our BRDF packaging. Update it to work with both variations.

Includes tests against both the new and old metadata.


geoscience-aman (approved):
Thanks Jeremy! Great that you added tests too! 🧪

PR #349, opened 2024-05-14
Make compatible with core 1.9

Ariana-B:
Update requirements to use datacube-core 1.9, replacing `datacube.config` with `datacube.cfg` and `datacube.utils.geometry` with `odc-geo`.


SpacemanPaul:
We will probably need to maintain parallel 1.8 and 1.9 branches for a while (as we are in core and OWS).  I don't think it's possible for one branch to work with both.

Ariana-B:
This needs to be kept as a parallel branch for now, so closing now that all tests are passing.

Issue #333, opened 2024-02-28
Formatting issue in docs


robbibt:
I don't know if this is a new issue, but recently I've found that the readthedocs formatting is a little broken, with API docs overlapping with the main content in the right column:

![image](https://github.com/opendatacube/eo-datasets/assets/17680388/cbb20dd9-6540-4b89-b1ac-1e4f77d4f754)

jeremyh:
Thanks for report. Odd, it looks like it reverted to the older readthedocs style when we fixed the build issue.

PR #354, opened 2024-06-25
Explicitly use the modern `sphinx_rtd_theme` doc theme

jeremyh:
[The docs](https://eodatasets.readthedocs.io/en/latest/) have reverted to a very old theme (I don't know why... did the default change?).

Explicitly set the modern rtd theme.

A demo build of this change is here: https://jez-eod3.readthedocs.io/en/latest/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,11 +58,4 @@ exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 # overwrite the builtin "default.css".
 # html_static_path = ["_static"]
 
-
-if is_on_readthedocs:
-    html_theme = "default"
-else:
-    import sphinx_rtd_theme
-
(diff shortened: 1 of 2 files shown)


jeremyh:
Previously reported by Robbi in #333

robbibt (approved):
Looks so much better! Only comment is that the content is currently a bit cramped on the left hand side, but not a major issue.

image

jeremyh:
> Looks so much better! Only comment is that the content is currently a bit cramped on the left hand side, but not a major issue.

That's odd that it's _slightly_ different to the old default...

jeremyh:
It's now the [latest](https://eodatasets.readthedocs.io/en/latest/) version of docs, but this wont, unfortunately, fix the previous tagged release build

PR #355, opened 2024-07-09
Ensure nodata=nan tags are written to floating point COGs.

### LINKED ISSUE #351: Important fix proposal: Implement setting "nan" nodata value for floating point rasters with no nodata attribute
Hi all, as discussed internally within DEA, we have recently encountered an issue where floating point rasters are not displayed correctly in ESRI software due to the lack of an explicitly set nodata attribute of "nan". This differs from open software such as QGIS and GDAL, which correctly interpret nodata by implicitly treating a missing nodata attribute as equivalent to "nan".

While this issue has been resolved by patching previously generated data on prod S3, a more sustainable long-term fix would be to set nodata attributes in our GeoTIFF writing code, to ensure that any floating point raster without a custom nodata attribute is assigned a nodata attribute of "nan". 

EO datasets is an important place to fix this, as it is used to generate derivative DEA products that often include floating point rasters. A possible fix would likely involve a minor change similar to psuedocode below:

```
if (raster is floating point dtype) AND (no nodata attribute set):
    raster.nodata = "nan"
```

Testing within DEA has revealed no downsteam impacts of this fix, while successfully solve the issue for ESRI users (a large proportion of our user base).

SpacemanPaul:
Basic attempt to fix #351 

Awaiting confirmation that COGs generated with this code work as expected with ArcGIS.

### DIFF
--- a/eodatasets3/images.py
+++ b/eodatasets3/images.py
@@ -712,7 +712,9 @@ class FileWrite:
         }
         # Ensure 'nan' is always tagged as nodata for floating point types
         if nodata is None:
-            nodata = float("nan") if numpy.issubdtype(array.dtype, numpy.floating) else 0
+            nodata = (
+                float("nan") if numpy.issubdtype(array.dtype, numpy.floating) else 0
+            )
 
         if nodata is not None:
             rio_args["nodata"] = nodata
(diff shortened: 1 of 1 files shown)


Ariana-B (approved):

robbibt (approved):
Works great for me in DEA Intertidal: 

`gdalinfo` on a floating point raster, before:
image

`gdalinfo` on a floating point raster, after:
image

PR #358, opened 2024-10-29
Integrate COG writing logic from odc-geo

Ariana-B:
As part of ODC 1.9 integration, we want to shift towards using `odc-geo` where possible.
Replace the COG writing code implemented here with the logic from `odc-geo` - this removes the need for the `FileWrite` and `GridSpec` classes.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -192,13 +192,13 @@ We can record them as source datasets, and the assembler can optionally copy any
 
 
 In these situations, we often write our new pixels as a numpy array, inheriting the existing
-:class:`grid spatial information <eodatasets3.GridSpec>` of our input dataset::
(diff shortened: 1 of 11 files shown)


omad:
Oh nice!

Core doesn't use `eo-datasets` though does it? 

So this isn't necessarily required for 1.9? but is likely to break downstream applications like `odc-stats`?

Ariana-B:
> Oh nice!
> 
> Core doesn't use `eo-datasets` though does it?
> 
> So this isn't necessarily required for 1.9? but is likely to break downstream applications like `odc-stats`?

Not required by core 1.9, no, just in line with the push in 1.9 to make sure we're using `odc-geo` instead of duplicating logic. This would indeed break `odc-stats`, but we should transition `odc-stats` to use `odc-geo` as a next step anyway.

PR #359, opened 2024-10-29
Remove distutils replace with shutils

alexgleith:
As of Python 3.12, `distutils` has been removed.

This change replaces the `distutils.spawn.find_executable` with `shutils.which`.

### DIFF
--- a/eodatasets3/verify.py
+++ b/eodatasets3/verify.py
@@ -3,7 +3,6 @@ import hashlib
 import logging
 import os
 import typing
-from distutils import spawn
 from pathlib import Path
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


eodatasets3/verify.py
```diff
@@ -35,11 +34,14 @@ def find_exe(name: str):
     :return: the absolute path to the executable.
     :rtype: str
     """
-    executable = spawn.find_executable(name)
-    if not executable:
-        raise Exception(f"No {name!r} command found.")
+    from shutil import which
```
omad, line 37: Probably should be a module level import like in the other PR.

omad (approved):
lgtm

PR #363, opened 2024-11-12
Add action to test build

Ariana-B:
Add job to test workflow to test build to avoid surprises at release publish when building for PyPI.
Also bump min python version to 3.10 for versioneer compatibility.


SpacemanPaul (approved):
LGTM

PR #373, opened 2025-02-05
Resolve numpy2 compatibility issues

Ariana-B:
Numpy was pinned to <2.0 due to an issue dealing with numpy floats in yaml serialisation - figured out cause and fix to make it numpy2 compatible.


eodatasets3/validate.py
```diff
@@ -615,7 +615,10 @@ def numpy_value_fits_dtype(value, dtype):
     if _is_nan(value):
         return np.issubdtype(dtype, np.floating)
     else:
-        return np.all(np.array([value], dtype=dtype) == [value])
+        try:
+            return np.all(np.array([value], dtype=dtype) == [value])
+        except Exception:  # pylint: disable=broad-except
```
omad, line 620: I guess this will work, but it seems a bit... heavy handed.

omad:
I think the new failures are due to breakages caused by the release of pytest 8+ (from last year some time). and pystac 1.12 a few weeks ago.

Also, the CI testing setup in here is painful, and the package setup is archaic. (Not at all on you!)

eodatasets3/validate.py
Ariana-B: I was getting an `OverflowError` in a test, but wasn't sure if that was the only kind of error that might be produced here.

Ariana-B:
@omad The datetime comparison causing the assertion errors comes from DeepDiff, not pytest. Some further digging reveals that ruamel-yaml >= 1.8.6 seems to be the culprit as tests pass locally with 1.8.5 and below, but I haven't yet figured out exactly what's going on - though I'm not sure how important this discrepancy is in the first place.

omad:
Merged into and resolved in #374

PR #374, opened 2025-02-06
Pin deps to fix tests, ditch docker for uv in GHA

omad:
I've broken my own rule and combined a few changes in the one PR, sorry!!

- This includes the numpy 2 fixes from #373 by @Ariana-B 
- And fixes deprecated usage of pytest.warn that breaks in pytest >= 8
- And fixes a couple of other warnings in the tests.
- And ditches `gdal` as a test dependency. According to the gdal docs, what we were testing for has been remedied since version 3.1 which came out 4 years ago. Overview block size now defaults to the same size as the main images instead of 128.
- By ditching `gdal` from testing, I can also ditch `docker` in GHA, which massively simplifies the test setup. And it's faster. Saves about 1m in GHA.
- i don't think eo-datasets needs it's own docker image. The DockerHub credentials seem to have been removed, which is lucky, because it was probably pushing GitHub tokens inside the image. So I've consolidated all the GitHub Actions workflows into one file.
- I'm also trialling running tests in parallel using `pytest-xdist`, which required removing a usage of `sys.argv` from one of the data prep scripts, which was only there to give a warning about outdated usage from years ago. Test execution in GHA drops from about 4m to 2.5m.
- We don't need to run pre-commit linting checks in GHA, they're being run by https://pre-commit-.ci/ . Which [is now the recommended approach](https://github.com/pre-commit/action).


robbibt:
@omad Not sure if it's useful, but re parallelization, we currently run our `eo-tides` `uv` tests in parallel directly via GitHub actions using a matrix strategy:

https://github.com/GeoscienceAustralia/eo-tides/blob/main/.github%2Fworkflows%2Fmain.yml#L40-L47

Ariana-B (approved):

omad:
> @omad Not sure if it's useful, but re parallelization, we currently run our `eo-tides` `uv` tests in parallel directly via GitHub actions using a matrix strategy:
> 
> https://github.com/GeoscienceAustralia/eo-tides/blob/main/.github%2Fworkflows%2Fmain.yml#L40-L47

Thanks Robbi, I always love looking at other people's GitHub Actions Workflows!

The GHA matrix is super useful for running tests in parallel in different environments (eg python versions). It could be used to split up a large and slow test suite with `pytest`, but you'd have to apply [marks](https://docs.pytest.org/en/stable/how-to/mark.html) or split up the suite some other way. 

I used [pytest-xdist](https://pypi.org/project/pytest-xdist/) here to run some of the tests in parallel in the same environment, without changing anything else. It works locally just as well as it works in CI. It is *super* sensitive to if there's any interdependencies with any of the tests. If they start randomly. failing, one test is relying on something happening in another test, and that's a bad thing.

PR #387, opened 2025-04-20
Fix some type signatures

pjonsson:


eodatasets3/properties.py
```diff
@@ -597,7 +600,7 @@ def producer(self, domain: str):
         self.properties["odc:producer"] = domain
 
     @property
-    def datetime_range(self) -> tuple[datetime, datetime]:
+    def datetime_range(self) -> tuple[datetime.datetime, datetime.datetime]:
```
pjonsson, line 603: I think this one broke in February this year, everything else has been wrong for a long time.

omad (approved):

PR #388, opened 2025-04-24
Use Dependabot for updates

pjonsson:
Make Dependabot provide the simple
updates so we don't have to do
it manually.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: docker
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/requirements"
+    schedule:
+      interval: "daily"
+    groups:
+      pip-deps:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: docker
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/requirements"
```
pjonsson, line 17: I searched a bit and the consensus seems to be that Dependabot will run some checks to identify the files to update, so I believe this should be enough for it to do the right thing.

pjonsson: Apparently I should have checked the next few links because there's always a stackoverflow thread with the answer: the files need to be called `<something>requirements.txt` (https://stackoverflow.com/questions/75899186/how-to-configure-dependabot-to-check-multiple-files).

Would you like me to rename the files to `deployment-requirements.txt`, `setup-requirements.txt' and `test-requirements.txt`? If you do, you can merge this PR as-is, and I will rename things in a separate PR tomorrow.

pjonsson: There is a list of the files Dependabot monitors under Insights->Dependency Graph (what?) in the Dependabot tab (https://github.com/opendatacube/datacube-core/network/updates/24378318/jobs). Clicking on the ".." next to `pyproject.toml` will list the files being monitored, so I guess we should merge this PR and then check if any files are missing.

omad (approved):

pjonsson:
All files in `requirements/` seem to be listed under https://github.com/opendatacube/eo-datasets/network/updates, so I'll assume this works as intended.

PR #397, opened 2025-05-02
Complete Ruff-migration

pjonsson:
Add Ruff as a dependency, remove the leftovers of black/isort, switch to Ruff from bandit, and remove the pep8-naming package+flake8 configuration since those rules are already enabled for Ruff.

Also remove some comments and some configuration directives that are no longer required.


omad (approved):
lgtm, thanks.

PR #398, opened 2025-05-02
dependabot: update boto* weekly instead

pjonsson:
Their release schedule is almost daily,
so update them once a week.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,11 +13,27 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
-  - package-ecosystem: "pip"
-    directory: "/requirements"
+  - package-ecosystem: "uv"
+    directory: "/"
     schedule:
       interval: "daily"
     groups:
-      pip-deps:
+      uv-deps:
         patterns:
           - "*"
+        exclude-patterns: # Packages updated almost every day.
+          - "boto3"
+          - "botocore"
+  - package-ecosystem: "uv"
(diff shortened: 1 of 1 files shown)


pjonsson:
This switches Dependabot to using the uv eco-system, but those still produce pull requests of the text files for some reason, so no loss in functionality from this PR.

I'm pretty sure Dependabot is broken for uv so I have prepared a reproducer that I haven't had time to bug report to them yet. Will do that during second half of next week.

pjonsson:
Judging from the PR submitted to datacube-core, it seems exclude did not work as I had understood it, so marking this as draft for now.

pjonsson:
I take it back, this PR should work:
* include "*" followed by exclude "something" results in everything except "something" being updated: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#patterns-and-exclude-patterns-groups
* the example here uses the same pattern: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates#example-2-grouped-updates-with-excluded-dependencies

pjonsson:
Yesterday I had https://github.com/opendatacube/datacube-core/network/updates/1009618341 in the dependabot status in datacube-core, today I have:
> Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'. Ecosystem 'uv' has overlapping directories.

which apparently can be worked around by setting target-branch for one of them, and letting the other use the default.

omad (approved):
👍🏻

PR #402, opened 2025-05-05
Unpin pystac

pjonsson:
Unpin pystac version and update
the expected file in tests for
the latest version. Add a version
constraint to the test groups so
that tests use a sufficiently
recent pystac version.

### DIFF
--- a/eodatasets3/properties.py
+++ b/eodatasets3/properties.py
@@ -313,6 +313,7 @@ _DERIVATIVE_PRODUCT_PROPS = {
 _STAC_MISC_PROPS = {
     "providers": None,  # https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#provider,
     # Projection extension
+    "proj:code": str,
     "proj:epsg": int,
     "proj:shape": None,
     "proj:transform": None,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
     "xarray",
     "datacube>=1.9.0",
     "python-rapidjson",
-    "pystac>=1.8.4,<1.12",
+    "pystac>=1.8.4",
 ]
 
 [project.urls]
@@ -67,6 +67,7 @@ all = [
     "deepdiff>=8.0",
     "h5py", # From wagl (and core)
     "mock",
+    "pystac>=1.12",
     "pytest-cov",
     "pytest-xdist",
     "pytest",
@@ -81,6 +82,7 @@ test = [
     "deepdiff>=8.0",
     "h5py", # From wagl (and core)
     "mock",
+    "pystac>=1.12",
     "pytest-cov",
     "pytest-xdist",
     "pytest",
--- a/requirements/deployment.txt
+++ b/requirements/deployment.txt
@@ -76,6 +76,8 @@ deprecat==2.1.3
     # via datacube
 distributed==2025.4.1
     # via datacube
+exceptiongroup==1.2.2
+    # via cattrs
 fiona==1.10.1
     # via eodatasets3 (pyproject.toml)
 fsspec==2025.3.2
@@ -86,6 +88,8 @@ greenlet==3.2.1
     # via sqlalchemy
 h5py==3.13.0
     # via eodatasets3 (pyproject.toml)
+importlib-metadata==8.7.0
+    # via dask
 jinja2==3.1.6
     # via distributed
 jmespath==1.0.1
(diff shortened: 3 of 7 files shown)


eodatasets3/properties.py
```diff
@@ -313,6 +313,7 @@ def parsed_sentinel_datastrip_id(tile_id) -> tuple[str, dict]:
 _STAC_MISC_PROPS = {
     "providers": None,  # https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#provider,
     # Projection extension
+    "proj:code": str,
```
pjonsson, line 316: I don't know anything about eo-datasets and how the code is supposed to work, I'm just guessing that `proj:code` should be alongside wherever `proj:epsg` was before.

pjonsson:
Pystac 1.12 didn't change any APIs, so this PR just makes the tests write/check STAC 1.1.0 files. Future releases of eo-datasets3 will write STAC 1.0.0 if pystac <1.12 is used, and STAC 1.1.0 when using pystac >= 1.12.

robbibt:
Hey @pjonsson, I believe this was pinned deliberately, because ODC tooling (particularly `odc-tools`) doesn't currently support STAC 1.1.0 metadata. This resulted in some very large continental runs at GA having to be re-run due to data being generated with incompatible output metadata.

I would personally love full STAC 1.1.0 support in ODC, and think that's a better solution going forward than pinning to `pystac<1.12`. But until that fix is made, undoing this pin would have downstream impacts.

I've been meaning to raise some issues about STAC 1.1.0 incompatiblity, so I might do that now and hopefully we can work out a plan for updating STAC support.

(ping @JM-GA who I think was the first to encounter the STAC 1.1.0 issue)

robbibt:
Issue here: https://github.com/opendatacube/odc-tools/issues/642

pjonsson:
@robbibt I realize this is probably some internal repository that was opensourced, but on a technical level: how is the STAC 1.1.0 support in odc-tools supposed to happen if eodatasets3 pins pystac on a version that doesn't have STAC 1.1.0 support?

robbibt:
> @robbibt I realize this is probably some internal repository that was opensourced, but on a technical level: how is the STAC 1.1.0 support in odc-tools supposed to happen if eodatasets3 pins pystac on a version that doesn't have STAC 1.1.0 support?

The main purpose of `eo-datasets` is to package up data with metadata that can be indexed into Open Data Cube. If the metadata produced by `eo-datasets` can't be indexed by Open Data Cube, that goes against the purpose of this repo. So I'm not suggesting we keep the pin forever - I just think the order of operations is should be to get full STAC 1.1.0 support in `odc-tools` first before we do so.

(I don't think @SpacemanPaul's [fix here](https://github.com/opendatacube/odc-tools/pull/643) itself has a dependency on `eo-datasets`, but I might be wrong)

pjonsson:
> [..] I just think the order of operations is should be to get full STAC 1.1.0 support in `odc-tools` first before we do so.
 
I've been up since 5:30 am this morning so it's not at all impossible I'm being dense, but how are you envisioning that to happen without this repository first unpinning pystac and making a release so the changes to odc-tools can be tested?

robbibt:
I also might be misinterpreting things, but STAC 1.1.0 support in `odc-tools` could be tested against any source of STAC 1.1.0 metadata - it doesn't need to be originally generated by `eo-datasets`.

pjonsson:
Sure, but odc-tools depend on eodatasets3:
```
odc-tools $ git grep eodatasets
...
apps/dc_tools/setup.cfg:    eodatasets3>=1.9
```
and eodatasets3 restricts `pystac>=1.8.4,<1.12`, so odc-tools will get pystac 1.11.0. Can pystac 1.11.0 deal with STAC 1.1.0 files without the PRs for STAC 1.1.0-support that were added in pystac 1.12.0 (https://github.com/stac-utils/pystac/releases/tag/v1.12.0)?

SpacemanPaul:
> Sure, but odc-tools depend on eodatasets3:
> 
> ```
> odc-tools $ git grep eodatasets
> ...
> apps/dc_tools/setup.cfg:    eodatasets3>=1.9
> ```
> 
> and eodatasets3 restricts `pystac>=1.8.4,<1.12`, so odc-tools will get pystac 1.11.0. Can pystac 1.11.0 deal with STAC 1.1.0 files without the PRs for STAC 1.1.0-support that were added in pystac 1.12.0 (https://github.com/stac-utils/pystac/releases/tag/v1.12.0)?

The issue with dc_tools is the code which converts from STAC to eo3 for indexing, which doesn't actually use pystac.  We can unpin here after we have merged the tools PR.

We should really backport to 1.8 as well.

omad (approved):
LGTM

PR #419, opened 2025-05-19
Fix tests in Py3.10. Use timezone.utc instead of datetime.UTC

omad:


omad:
OMG, Python 3.10 is surprisingly horrible.

PR #458, opened 2025-08-18
readthedocs: use uv for building

pjonsson:
This uses uv.lock for the dependencies,
which eliminates the need for an extra
requirements.txt in docs/.


pjonsson:
Apparently the documentation isn't built on pull requests here, but this is the `.readthedocs.yml` from explorer but updated to use the test group that contains the doc dependencies in this repository.

omad (approved):

PR #490, opened 2025-11-02
Add `dea_versioned` naming conventions

jeremyh:
Adds options `conventions="dea_versioned"` and `conventions="dea_s2_versioned"`.

It would be nice to combine them into one naming convention, but that would require a bigger refactor of eodatasets: they could no longer be static, as the conventions would have to change slightly depending on satellite.

### DIFF
--- a/eodatasets3/names.py
+++ b/eodatasets3/names.py
@@ -930,6 +930,14 @@ class DEAS2DerivativesNamingConventions(DEADerivativesNamingConventions):
         )
 
 
+class DEAVersionedNamingConventions(DEANamingConventions):
+    """
+    TODO
+    """
+
+    ...
+
+
 class AfricaProductName:
     def __get__(self, c: "NamingConventions", owner) -> str:
         if c.metadata.product_name:
@@ -976,6 +984,7 @@ class DEAfricaNamingConventions(NamingConventions):
 KNOWN_CONVENTIONS = dict(
     default=NamingConventions,
     dea=DEANamingConventions,
+    dea_versioned=DEAVersionedNamingConventions,
     dea_s2=DEAS2NamingConventions,
     dea_s2_derivative=DEAS2DerivativesNamingConventions,
     dea_c3=DEADerivativesNamingConventions,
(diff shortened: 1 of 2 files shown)


jeremyh:
@robbibt  I'm currently calling them `dea_versioned` and `dea_s2_versioned` but other suggestions are welcome :)

sixy6e:
> It would be nice to combine them into one naming convention, but that would require a bigger refactor of eodatasets: they could no longer be static, as the conventions would have to change slightly depending on satellite.

Yeah, it would be nice, and the only reason S2 is different is to account for the insertion of the **_datatake sensing time_** that already exists in GAs data products.

In a future state, consolidation would be ideal.

But for now, as it's explicit for the S2 case, I'd suggest the same for Landsat, i.e `dea_ls_versioned`.  I'd recommend dropping the `a` in `dea` though.  How much is there in the existing codebase that refers to `dea`?

eodatasets3/names.py
```diff
@@ -930,6 +940,26 @@ def __init__(self, properties: Mapping) -> None:
         )
 
 
+class DEAVersionedNamingConventions(DEANamingConventions):
```
sixy6e, line 943: Is there any benefit to being explicit with a `DEALandsatVersionedNamingConventions` as what is done for `DEAS2VersionedNamingConventions`?  Or simply that this would be generic that items other and Landsat could use it?

jeremyh: Yes, in theory they were designed to be multi-sensor originally, it's just s2 that broke it. They're still intended to be the general defaults when a platform doesn't need its own unique conventions.

It's also for consistency with the other naming conventions in the codebase. We could change them to "de" as mentioned above, though it might sound odd for there to only be one set called "de"

sixy6e (approved):
Everything looks good. Sorry, missed the notiifcations regarding my Q's.

jeremyh:
Adding second reviewer as it doesn't allow merge yet

pjonsson (approved):
I don't know this stuff but it looks like the PR should be merged, so I'll press the button.

PR #522, opened 2026-01-07
Handle updated metadata format for BRDF

jeremyh:
Update the code to support both old and new.

There are no existing tests in this repo for this module (and it hasn't been touched in a long time): but it has successfully run on both versions of datasets.


SpacemanPaul (approved):
I'm not very familiar with this code or this metadata, but the PR seems reasonable enough to me.

Would be good to add test case though - even a partial one would be better than nothing.

uchchwhash (approved):
Looks good, also, worked for me.

Cheers!

